PR #1, opened 2019-12-19
add "/debug" endpoint for pprof support

clamoriniere:
This PR introduce the possibility to enable `pprof` behind the "/debug" endpoint in the extendeddaemonset endpoint.

* add args --pprof to activate `/debug` endpoint (false by default)
* add args --pprof-bind-address to list on a specific address and port
* remove the metrics k8s service creation

### DIFF
--- a/pkg/controller/debug/register.go
+++ b/pkg/controller/debug/register.go
@@ -11,7 +11,7 @@ import (
 	"github.com/datadog/extendeddaemonset/pkg/controller/httpserver"
 )
 
-// Options use to provide configuration option
+// Options used to provide configuration options
 type Options struct {
 	CmdLine bool
 	Profile bool
--- a/pkg/controller/debug/register.go
+++ b/pkg/controller/debug/register.go
@@ -29,7 +29,7 @@ func DefaultOptions() *Options {
 	}
 }
 
-// Register use to register the different debug endpoint
+// Register used to register the different debug endpoints
 func Register(mux httpserver.Server, options *Options) {
 	mux.HandleFunc("/debug/pprof/", pprof.Index)
 	if options == nil {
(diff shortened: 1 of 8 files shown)


pkg/controller/debug/register.go
```diff
@@ -0,0 +1,50 @@
...
+
+package debug
+
+import (
+	"net/http/pprof"
+
+	"github.com/datadog/extendeddaemonset/pkg/controller/httpserver"
+)
+
+// Options use to provide configuration option
```
ahmed-mez:
```suggestion
// Options uses to provide configuration options
```

pkg/controller/debug/register.go
```diff
@@ -0,0 +1,50 @@
...
+func DefaultOptions() *Options {
+	return &Options{
+		CmdLine: true,
+		Profile: true,
+		Symbol:  true,
+		Trace:   true,
+	}
+}
+
+// Register use to register the different debug endpoint
```
ahmed-mez:
```suggestion
// Register used to register the different debug endpoints
```

pkg/controller/httpserver/runner.go
```diff
@@ -0,0 +1,101 @@
...
+	"net/http"
+
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var log = logf.Log.WithName("httpserver")
+
+// DefaultBindAddress sets the default bind address for the HTTP server listener
+var DefaultBindAddress = ":8080"
```
ahmed-mez: I don't see any reason why this is public

clamoriniere: if they want to use it as default args value

PR #5, opened 2020-03-02
Add EDS resources metrics

clamoriniere:
* Add equivalent to KSM metrics for ExtendedDaemonSet resource
* Update operator-sdk version to `0.15.2`
* Add CRD `clientset`,`listener` generation


pkg/controller/extendeddaemonset/metrics.go
```diff
@@ -0,0 +1,293 @@
...
+	resourceNamePromLabel      = "name"
+	resourceNamespacePromLabel = "namespace"
+
+	extendeddaemonsetCreated                        = "extendeddaemonset_created"
+	extendeddaemonsetStatusDesired                  = "extendeddaemonset_status_desired"
+	extendeddaemonsetStatusCurrent                  = "extendeddaemonset_status_current"
+	extendeddaemonsetStatusReady                    = "extendeddaemonset_status_ready"
+	extendeddaemonsetStatusAvailable                = "extendeddaemonset_status_available"
+	extendeddaemonsetStatusUpToDate                 = "extendeddaemonset_status_uptodate"
+	extendeddaemonsetStatusIgnoredUnresponsiveNodes = "extendeddaemonset_status_ignored_unresponsive_nodes"
```
ahmed-mez: shouldn't we use `eds_` as a prefix instead to avoid getting very long metric names?

pkg/controller/extendeddaemonset/metrics.go
```diff
@@ -0,0 +1,293 @@
...
+							LabelValues: getLabelsValues(eds),
+						},
+					},
+				}
+			},
+		},
+		{
+			Name: extendeddaemonsetStatusIgnoredUnresponsiveNodes,
+			Type: ksmetric.Gauge,
+			Help: "The total number of nodes that are ignore by the rolling update strategy due to an unresponsive state",
```
ahmed-mez:
```suggestion
			Help: "The total number of nodes that are ignored by the rolling update strategy due to an unresponsive state",
```

PR #10, opened 2020-04-19
Update Pod if Node annotation was updated

clamoriniere:


pkg/controller/extendeddaemonsetreplicaset/strategy/rollingupdate.go
```diff
@@ -53,7 +53,7 @@ func ManageDeployment(client client.Client, params *Parameters) (*Result, error)
 				nbIgnoredUnresponsiveNodes++
 				continue
 			}
-			if !compareSpecTemplateMD5Hash(params.Replicaset.Spec.TemplateGeneration, pod) {
+			if !compareCurrentPodWithNewPod(params.EDSName, params.Replicaset, pod, node) {
```
ahmed-mez, line 56: should we make  `compareCurrentPodWithNewPod` accept `Parameters` as an argument here instead of `params.EDSName` and `params.Replicaset` ? 
```suggestion
			if !compareCurrentPodWithNewPod(params, pod, node) {
```

pkg/controller/extendeddaemonsetreplicaset/strategy/utils.go
```diff
@@ -23,9 +23,29 @@ import (
 	datadoghqv1alpha1 "github.com/datadog/extendeddaemonset/pkg/apis/datadoghq/v1alpha1"
 	"github.com/datadog/extendeddaemonset/pkg/controller/extendeddaemonsetreplicaset/conditions"
 	podaffinity "github.com/datadog/extendeddaemonset/pkg/controller/utils/affinity"
+	"github.com/datadog/extendeddaemonset/pkg/controller/utils/comparison"
 	podutils "github.com/datadog/extendeddaemonset/pkg/controller/utils/pod"
 )
 
+func compareCurrentPodWithNewPod(edsName string, replicaset *datadoghqv1alpha1.ExtendedDaemonSetReplicaSet, pod *corev1.Pod, node *corev1.Node) bool {
+	// check that the pod correspond to the replicaset. if not return false
+	if !compareSpecTemplateMD5Hash(replicaset.Spec.TemplateGeneration, pod) {
```
ahmed-mez, line 32: should we check for nil pointers here ?

pkg/controller/extendeddaemonsetreplicaset/strategy/utils.go
```diff
@@ -23,9 +23,29 @@ import (
 	datadoghqv1alpha1 "github.com/datadog/extendeddaemonset/pkg/apis/datadoghq/v1alpha1"
 	"github.com/datadog/extendeddaemonset/pkg/controller/extendeddaemonsetreplicaset/conditions"
 	podaffinity "github.com/datadog/extendeddaemonset/pkg/controller/utils/affinity"
+	"github.com/datadog/extendeddaemonset/pkg/controller/utils/comparison"
 	podutils "github.com/datadog/extendeddaemonset/pkg/controller/utils/pod"
 )
 
+func compareCurrentPodWithNewPod(edsName string, replicaset *datadoghqv1alpha1.ExtendedDaemonSetReplicaSet, pod *corev1.Pod, node *corev1.Node) bool {
+	// check that the pod correspond to the replicaset. if not return false
```
ahmed-mez, line 31: nit
```suggestion
	// check that the pod corresponds to the replicaset. if not return false
```

pkg/controller/extendeddaemonsetreplicaset/strategy/utils.go
```diff
@@ -23,9 +23,29 @@ import (
...
+	if !compareSpecTemplateMD5Hash(replicaset.Spec.TemplateGeneration, pod) {
+		return false
+	}
+	if !compareNodeResourcesAnnotationsMD5Hash(edsName, replicaset, pod, node) {
+		return false
+	}
+	return true
+}
+
+func compareNodeResourcesAnnotationsMD5Hash(edsName string, replicaset *datadoghqv1alpha1.ExtendedDaemonSetReplicaSet, pod *corev1.Pod, node *corev1.Node) bool {
```
ahmed-mez, line 41: same question, is there a risk of receiving nil pointers here ?

pkg/controller/utils/comparison/comparison.go
```diff
@@ -73,3 +74,26 @@ func StringsContains(a []string, x string) bool {
...
+func GenerateHashFromNodeAnnotation(edsNamespace, edsName string, nodeAnnotations map[string]string) string {
+	// build prefix for this specific eds
+	prefixKey := fmt.Sprintf(datadoghqv1alpha1.ExtendedDaemonSetRessourceNodeAnnotationKey, edsNamespace, edsName, "")
+
+	resourcesAnnotations := []string{}
+	for key, value := range nodeAnnotations {
+		if strings.HasPrefix(key, prefixKey) {
+			resourcesAnnotations = append(resourcesAnnotations, fmt.Sprintf("%s=%s", key, value))
+		}
+	}
```
ahmed-mez, line 88: nit, we can hash in this block and avoid allocating `resourcesAnnotations`

pkg/controller/utils/comparison/comparison.go
```diff
@@ -73,3 +74,26 @@ func StringsContains(a []string, x string) bool {
 	}
 	return false
 }
+
+// GenerateHashFromNodeAnnotation use to generate the MD5 hash from Node annotations
```
ahmed-mez, line 78: nit
```suggestion
// GenerateHashFromNodeAnnotation is used to generate the MD5 hash from Node annotations
```

pkg/controller/utils/comparison/comparison.go
```diff
@@ -73,3 +74,26 @@ func StringsContains(a []string, x string) bool {
...
+func GenerateHashFromNodeAnnotation(edsNamespace, edsName string, nodeAnnotations map[string]string) string {
```
ahmed-mez, line 79: I find the function name confusing, the name imply that we're hashing all the node annotations but in fact we're only hashing annotations with `ExtendedDaemonSetRessourceNodeAnnotationKey` prefix. Can we adapt the function name or at least describe this in the function documentation?

ahmed-mez:
nice PR, added few comments - could you please add a unit test as well? 🙇

pkg/controller/extendeddaemonsetreplicaset/strategy/utils.go
clamoriniere: on the `replicaset`? It is a mandatory args. and also the function is private. So I think it is fine to not check if it nil. wdyt?

pkg/controller/extendeddaemonsetreplicaset/strategy/utils.go
clamoriniere: I don't think so

pkg/controller/utils/comparison/comparison.go
clamoriniere: I didn't done it in this loop, because the order in `range map` is not always the same.

other implementation that I can do: store and sort only the keys in a slice, and loop a second time on the map? do yo think it will be a big win?

pkg/controller/extendeddaemonsetreplicaset/strategy/utils.go
ahmed-mez: okay

pkg/controller/utils/comparison/comparison.go
ahmed-mez: okay but we're not sorting `resourcesAnnotations` before calculating the hash and we're populating the slice with the random order of the map. I think we need to sort the slice before hashing

clamoriniere: I remove the line by mistake :)
pr updated

PR #14, opened 2020-04-29
Avoid adding new annotation if not needed

clamoriniere:
avoid adding "MD5NodeExtendedDaemonSetAnnotationKey" annotation if the value is empty


pkg/controller/extendeddaemonsetreplicaset/strategy/utils.go
```diff
@@ -44,7 +44,7 @@ func compareCurrentPodWithNewPod(params *Parameters, pod *corev1.Pod, node *Node
 
 func compareNodeResourcesOverwriteMD5Hash(edsName string, replicaset *datadoghqv1alpha1.ExtendedDaemonSetReplicaSet, pod *corev1.Pod, node *NodeItem) bool {
 	nodeHash := comparison.GenerateHashFromEDSResourceNodeAnnotation(replicaset.Namespace, edsName, node.Node.GetAnnotations())
-	if val, ok := pod.Annotations[datadoghqv1alpha1.MD5NodeExtendedDaemonSetAnnotationKey]; ok && val == nodeHash {
+	if val, ok := pod.Annotations[datadoghqv1alpha1.MD5NodeExtendedDaemonSetAnnotationKey]; nodeHash == "" || ok && val == nodeHash {
```
L3n41c, line 47: To consider that the annotation hasn’t changed compared to when the pod was created, when the node has no annotation, we must check that the pod doesn’t have the checksum annotation.

```suggestion
	if val, ok := pod.Annotations[datadoghqv1alpha1.MD5NodeExtendedDaemonSetAnnotationKey]; !ok && nodeHash == "" || ok && val == nodeHash {
```

PR #17, opened 2020-05-13
Update operator-sdk to 0.17.0 and fix issues for Kube 1.18

vboulineau:
Important note: when upgrading to this version, EDS must define the "protocol" attribute, will be considered invalid otherwise


chart/extendeddaemonset/templates/deployment.yaml
```diff
@@ -31,6 +31,8 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
             - --zap-level={{ .Values.logLevel }}
+            - --zap-encoder=console
```
clamoriniere, line 34: could you add those settings in the `/deploy/operator.yaml` too

PR #22, opened 2020-06-12
Enable the possiblity to pause a Canary deployment

juliogreff:
In the ExtendedDaemonset.Spec.Strategy, it is possible to define a
duration for the Canary deployment. At the end of this duration, if the
ExtendedDaemonset was not rolled back, the Canary deployment becomes
active and the rolling-update continues on all the nodes. By adding a
Paused field, it's now possible to prevent that from happening until
Paused is set back to false again. That's useful to investigate faulty
deployments as they currently are, without the need for a full rollback.

Another way we could've implemented this would be with an annotation,
but the last thing one needs when trying to prevent disaster is
misspelling an annotation that's not validated, so I preferred to go
with a new field.

### DIFF
--- a/pkg/plugin/pause.go
+++ b/pkg/plugin/pause.go
@@ -20,10 +20,15 @@ import (
 	"github.com/datadog/extendeddaemonset/pkg/apis/datadoghq/v1alpha1"
 )
 
+const (
+	cmdPause   = true
+	cmdUnpause = false
+)
+
 var (
 	pauseExample = `
-	# pause a canary replicaset
-	%[1]s pause foo
+	# %[1]s a canary deployment
+	kubectl eds %[1]s foo
 `
 )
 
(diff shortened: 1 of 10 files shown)


pkg/plugin/pause.go
```diff
@@ -0,0 +1,147 @@
...
+		return fmt.Errorf("unable to get ExtendedDaemonSet, err: %v", err)
+	}
+
+	if eds.Spec.Strategy.Canary == nil {
+		return fmt.Errorf("the ExtendedDaemonset does not have a canary")
+	}
+
+	if eds.Spec.Strategy.Canary.Paused {
+		return fmt.Errorf("ExtendedDaemonset '%s/%s' deployment already paused", o.userNamespace, o.userExtendedDaemonSetName)
+	}
```
ahmed-mez, line 135: Can we implement an unpause command ?

I'm thinking we can refactor a bit what we have currently and implement a new command `canary` with subcommands `validate` `pause` and `unpause`. I know this is extra work and wasn't in the scope of the task though. We can do it later.
cc @clamoriniere in case you have something to add here

juliogreff: I don't mind expanding the scope here, it's a relatively small change. I'll add the subcommand.

ahmed-mez: thank you 🙇

ahmed-mez: here is an example on how we handle subcommands in the [operator kubectl-plugin ](https://github.com/DataDog/datadog-operator/blob/master/cmd/kubectl-datadog/agent/agent/agent.go#L37-L39)

ahmed-mez: also [here](https://github.com/DataDog/datadog-operator/blob/master/cmd/kubectl-datadog/datadog/datadog.go#L39-L48)

PR #25, opened 2020-07-17
Add some facilities to tune the selection of canary nodes

L3n41c:
This PR adds two features to select canary nodes in a more fine-grained way:
* a node selector to choose the canary nodes: only the nodes matching the selector will be eligible to be used as canary.
  Ex. of use:
  ```yaml
  spec:
    strategy:
      canary:
        nodeSelector:
          matchExpressions:
          - key: canary
            operator: In
            values:
            - "true"
  ```
* a node anti-affinity feature: given a set of label keys, the canary nodes will be chosen so that the corresponding label values are different.
  Ex. of use:
  ```yaml
  spec:
    strategy:
      canary:
        nodeAntiAffinityKeys:
        - service
  ```

### DIFF
--- a/pkg/controller/extendeddaemonset/extendeddaemonset_controller.go
+++ b/pkg/controller/extendeddaemonset/extendeddaemonset_controller.go
@@ -291,7 +291,7 @@ func (r *ReconcileExtendedDaemonSet) selectNodes(logger logr.Logger, daemonsetSp
 
 	listOptions := []client.ListOption{}
 	if replicaset.Spec.Selector != nil {
-		selector, err := utils.LabelSelector2LabelSelector(logger, replicaset.Spec.Selector)
+		selector, err := utils.ConvertLabelSelector(logger, replicaset.Spec.Selector)
 		if err != nil {
 			logger.Error(err, "Failed to parse label selector")
 		} else {
@@ -301,7 +301,7 @@ func (r *ReconcileExtendedDaemonSet) selectNodes(logger logr.Logger, daemonsetSp
 		}
 	}
 	if daemonsetSpec.Strategy.Canary.NodeSelector != nil {
-		selector, err := utils.LabelSelector2LabelSelector(logger, daemonsetSpec.Strategy.Canary.NodeSelector)
+		selector, err := utils.ConvertLabelSelector(logger, daemonsetSpec.Strategy.Canary.NodeSelector)
 		if err != nil {
 			logger.Error(err, "Failed to parse label selector")
 		} else {
@@ -352,14 +352,14 @@ func (r *ReconcileExtendedDaemonSet) selectNodes(logger logr.Logger, daemonsetSp
 		if len(daemonsetSpec.Strategy.Canary.NodeAntiAffinityKeys) != 0 {
 			for _, node := range nodeList.Items {
 
-				antiAffinityKeysValueString := antiAffinityKeysValue(&node, daemonsetSpec)
-				if _, found := antiAffinityKeysValues[antiAffinityKeysValueString]; !found {
-					antiAffinityKeysValues[antiAffinityKeysValueString] = 0
+				antiAffinityKeysValue := getAntiAffinityKeysValue(&node, daemonsetSpec)
+				if _, found := antiAffinityKeysValues[antiAffinityKeysValue]; !found {
+					antiAffinityKeysValues[antiAffinityKeysValue] = 0
 				}
 
 				for _, currentNode := range currentNodes {
 					if node.Name == currentNode {
-						antiAffinityKeysValues[antiAffinityKeysValueString]++
+						antiAffinityKeysValues[antiAffinityKeysValue]++
 						break
 					}
 				}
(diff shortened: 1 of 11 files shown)


pkg/controller/extendeddaemonsetreplicaset/ddaemonsetreplicaset_controller.go
```diff
@@ -419,15 +425,20 @@ func (r *ReconcileExtendedDaemonSetReplicaSet) getOldDaemonsetPodList(ds *datado
...
-	var podSelector labels.Set
+	podListOptions := []client.ListOption{}
 	if oldDaemonset.Spec.Selector != nil {
-		podSelector = labels.Set(oldDaemonset.Spec.Selector.MatchLabels)
-	}
-	podListOptions := []client.ListOption{
-		client.MatchingLabelsSelector{
-			Selector: podSelector.AsSelectorPreValidated(),
-		},
+		selector, err2 := utils.LabelSelector2LabelSelector(log, oldDaemonset.Spec.Selector)
```
celenechang, line 430: is the error var name intentional?

pkg/controller/utils/labelselector.go
```diff
@@ -0,0 +1,50 @@
...
+			req, err2 := labels.NewRequirement(expr.Key, op, expr.Values)
+			if err2 != nil {
+				logger.Error(err, "NewRequirement")
+				err = err2
+				continue
+			}
+			outSelector = outSelector.Add(*req)
+		}
+	}
+	return outSelector, err
```
celenechang, line 49: it looks like even though we continue through the loops when there is an error here, we `return` on error when `LabelSelector2LabelSelector` is called (e.g. https://github.com/DataDog/extendeddaemonset/pull/25/files#diff-6d4acc3f7e4675ba93d62c645e490c0eR370) or we don't use the returned selector (https://github.com/DataDog/extendeddaemonset/pull/25/files?file-filters%5B%5D=.go#diff-6f8ca7266c71f02fdcdcc35fb5f12acaR294). is it worth continue the loop in these cases?

pkg/controller/utils/labelselector.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"github.com/go-logr/logr"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+)
+
+// LabelSelector2LabelSelector converts a "k8s.io/apimachinery/pkg/apis/meta/v1".LabelSelector as found in manifests spec section into a "k8s.io/apimachinery/pkg/labels".Selector to be used to filter list operations.
+func LabelSelector2LabelSelector(logger logr.Logger, inSelector *metav1.LabelSelector) (outSelector labels.Selector, err error) {
```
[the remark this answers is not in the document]
L3n41c: That’s smarter, indeed.
Name changed !

pkg/controller/extendeddaemonset/extendeddaemonset_controller.go
```diff
@@ -307,30 +323,79 @@ func (r *ReconcileExtendedDaemonSet) selectNodes(logger logr.Logger, daemonsetSp
...
+
+	// Look for other nodes to use as canary
 	if len(currentNodes) < nbCanaryPod {
+		antiAffinityKeysValues := make(map[string]int)
+
+		// Look for the values of the labels set as NodeAntiAffinityKeys of the nodes already selected as canary
+		if len(daemonsetSpec.Strategy.Canary.NodeAntiAffinityKeys) != 0 {
+			for _, node := range nodeList.Items {
+
+				antiAffinityKeysValueString := antiAffinityKeysValue(&node, daemonsetSpec)
```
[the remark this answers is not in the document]
L3n41c: Indeed.

pkg/controller/extendeddaemonset/extendeddaemonset_controller.go
```diff
@@ -307,30 +323,79 @@ func (r *ReconcileExtendedDaemonSet) selectNodes(logger logr.Logger, daemonsetSp
...
+			if alreadySelected {
 				continue
 			}
-			if !found {
-				currentNodes = append(currentNodes, node.Name)
+
+			// Check if the node labels selected by `NodeAntiAffinityKeys` are different from the ones of the already selected canary nodes.
+			if len(daemonsetSpec.Strategy.Canary.NodeAntiAffinityKeys) != 0 {
+				antiAffinityKeysValueString := antiAffinityKeysValue(&node, daemonsetSpec)
+				if nb := antiAffinityKeysValues[antiAffinityKeysValueString]; nb >= (nbCanaryPod+len(antiAffinityKeysValues)-1)/len(antiAffinityKeysValues) {
```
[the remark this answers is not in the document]
L3n41c: I’ve put the rational of this expression in a comment directly inside the code.

pkg/controller/extendeddaemonsetreplicaset/ddaemonsetreplicaset_controller.go
L3n41c: It was to fix a linter error complaining about variable shadowing.
But it’s clearer if I remove the named output parameter.
So, I changed to use `err` now.

pkg/controller/utils/labelselector.go
L3n41c: You’re right.
I suppose it’s clearer to generate an error rather than silently ignoring only the part of a selector which is not valid.
I’ve changed to use early `return`s.

pkg/controller/extendeddaemonset/extendeddaemonset_controller.go
```diff
@@ -380,13 +380,26 @@ func (r *ReconcileExtendedDaemonSet) selectNodes(logger logr.Logger, daemonsetSp
...
+			// In our example, that map would have two entries: `A` and `B`.
+			// We want a maximum of `nb_canaries / nb_different_values` (4/2=2) for each value.
+			// So, we want to reject a node if it would make the selection unbalanced, i.e. if
+			// antiAffinityKeysValues[getAntiAffinityKeysValue(&node, daemonsetSpec)] >= nbCanaryPod / len(antiAffinityKeysValues)
+			//
+			// In the above mathematical expression, we want the division result to be rounded up.
+			// In our example, if we want 5 canary nodes, we will want 3 nodes labeled `service=A` and 2 nodes labeled `service=B` or the other way round
+			// so, we want to reject nodes as soon as the number of selected nodes with that label value exceeds 3 = ceil(5/2)
+			//
+			// An efficient way to compute `ceil(a/b)` with only integer computing is to compute `(a+b-1)/b`.
```
celenechang, line 396: 👏 x 💯

pkg/controller/extendeddaemonset/extendeddaemonset_controller.go
```diff
@@ -352,14 +352,14 @@ func (r *ReconcileExtendedDaemonSet) selectNodes(logger logr.Logger, daemonsetSp
 		if len(daemonsetSpec.Strategy.Canary.NodeAntiAffinityKeys) != 0 {
 			for _, node := range nodeList.Items {
 
-				antiAffinityKeysValueString := antiAffinityKeysValue(&node, daemonsetSpec)
-				if _, found := antiAffinityKeysValues[antiAffinityKeysValueString]; !found {
-					antiAffinityKeysValues[antiAffinityKeysValueString] = 0
+				antiAffinityKeysValue := getAntiAffinityKeysValue(&node, daemonsetSpec)
```
celenechang, line 355: nit suggestion: perhaps for clarity, we could rename the `antiAffinityKeysValues` map to something less similar to `antiAffinityKeysValue` ?

PR #27, opened 2020-07-24
Document pod exclusion on a node

ahmed-mez:


README.md
```diff
@@ -208,6 +208,34 @@ spec:
...
+
+```yaml
+apiVersion: datadoghq.com/v1alpha1
+kind: ExtendedDaemonSet
+metadata:
+  name: foo
+spec:
+  selector:
+    matchExpressions:
+      - {key: extendeddaemonset.datadoghq.com/exclude, operator: NotIn, values: [foo]}
```
L3n41c, line 225: Does it work to use `operator: DoesNotExist` without any `values` ?
It would allow us to not care on the value.

The drawback would be that a node labelled with `…/exclude=false` would be excluded as well.

I’m wondering what would be the more natural setup.

ahmed-mez: I think the value matters because we can have multiple EDS running, with the value we guarantee the node label corresponds to the right EDS (the value equals the EDS name,  `foo` in this example)

PR #28, opened 2020-07-28
add feature to autopause canary

celenechang:
- updates pause feature to use annotation over spec field (to prevent churning of resource)
- adds autopause feature which will pause a canary deployment if restarts are detected for a canary pod, and give the `reason` from the pod with the most restarts
- adds autopause `reason` to the EDS status
- merges `Canary Paused` into the existing `Status` output, for clarity. (next PR will also add `Canary Failed` status option)
- adds unit tests
- adds TODO for e2e test (ran into blocking issues in development)


pkg/plugin/get.go
```diff
@@ -154,7 +154,7 @@ func (o *GetOptions) Run() error {
 
 func newGetTable(out io.Writer) *tablewriter.Table {
 	table := tablewriter.NewWriter(out)
-	table.SetHeader([]string{"Namespace", "Name", "Desired", "Current", "Ready", "Up-to-date", "Available", "Ignored Unresponsive Nodes", "Status", "Active RS", "Canary RS", "Canary Paused", "Age"})
+	table.SetHeader([]string{"Namespace", "Name", "Desired", "Current", "Ready", "Up-to-date", "Available", "Ignored Unresponsive Nodes", "Status", "Reason", "Active RS", "Canary RS", "Age"})
```
juliogreff, line 157: This change needs to be reflected in the CRD as well: https://github.com/DataDog/extendeddaemonset/blob/master/pkg/apis/datadoghq/v1alpha1/extendeddaemonset_types.go#L126-L136

PR #29, opened 2020-07-30
Add ability to fail a canary

celenechang:
This PR allows a Canary deployment to be manually failed using the `kubectl-eds` plugin: `kubectl-eds canary fail <eds name>`.

This will result in 
- the Canary deployment being cancelled
- the currently active replicaset remaining active

Note that the EDS status will show as `Canary Failed` until it is reset with `kubectl-eds canary reset <eds name>`. This is intended to help prevent a silent failure if another Canary deployment is attempted while the `failed` annotation is still set to `"true"`.


pkg/apis/datadoghq/v1alpha1/extendeddaemonset_types.go
```diff
@@ -80,18 +80,20 @@ type ExtendedDaemonSetSpecStrategyCanary struct {
 	NodeAntiAffinityKeys []string `json:"nodeAntiAffinityKeys,omitempty"`
 	// +optional
 	Paused bool `json:"paused,omitempty"`
+	// +optional
+	Failed bool `json:"failed,omitempty"`
```
L3n41c, line 84: Is this `spec.canary` field used ?
I have the feeling that the status of the EDS is stored in `status.state` and the channel used by a user to force a failed state is annotations rather than this `spec` field.

celenechang: it's not! this is a relic from the first version, thanks for pointing it out

PR #30, opened 2020-07-30
Release process

L3n41c:


deploy/olm-catalog/extendeddaemonset/manifests/datadoghq.com_extendeddaemonsetreplicasets_crd.yaml
```diff
@@ -0,0 +1,6090 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
clamoriniere, line 1: should we have a copy of the CRD in this folder? can we do a symlink?

L3n41c: If I remember well, I tried to use a symlink here and it broke something.

clamoriniere (approved):

PR #36, opened 2020-09-04
Move the CRDs of the chart to a dedicated directory

L3n41c:
https://helm.sh/docs/chart_best_practices/custom_resource_definitions/


clamoriniere (commented):
Do it works for helm2?

If not, we should add a readme.md in chart to explain the limitation: Helm2 not supported

clamoriniere (approved):

PR #41, opened 2020-09-09
Fix exclude nodes feature

clamoriniere:
* Update documentation: how to exclude temporary nodes
* Fix the `selectNodes()` function that was using the Pod labelSelector
  to filter Nodes.
* Add some logic in the function that inject the NodeName field affinity
  to merge the NodeAffinity instead of replacing it.

## Test plan:
* deploy the `examples/foo-eds_exclude-node.yaml` EDS in a test cluster.
* wait to have all the daemonset pod created
* add the label on a node. for example: `kubectl label node kind-worker extendeddaemonset.datadoghq.com/exclude=foo --overwrite`
* the daemonset pod on the labeled node should be deleted
* remove the label`kubectl label node kind-worker extendeddaemonset.datadoghq.com/exclude- --overwrite`, the daemonset pod should be reschedule on the node.

### DIFF
--- a/pkg/controller/utils/affinity/affinity.go
+++ b/pkg/controller/utils/affinity/affinity.go
@@ -54,7 +54,7 @@ func ReplaceNodeNameNodeAffinity(affinity *v1.Affinity, nodename string) *v1.Aff
 		return affinity
 	}
 
-	// Replace node selector with the new one.
+	// Build new NodeSelectorTerm list to add the Node name field selector
 	newSelectorTerms := []v1.NodeSelectorTerm{}
 	for _, term := range nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms {
 		newTerm := term.DeepCopy()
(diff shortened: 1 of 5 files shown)


pkg/controller/utils/affinity/affinity.go
```diff
@@ -55,11 +55,27 @@ func ReplaceNodeNameNodeAffinity(affinity *v1.Affinity, nodename string) *v1.Aff
...
-			MatchFields: []v1.NodeSelectorRequirement{nodeSelReq},
-		},
+	newSelectorTerms := []v1.NodeSelectorTerm{}
+	for _, term := range nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms {
+		newTerm := term.DeepCopy()
+
+		if len(newTerm.MatchFields) == 0 {
+			newTerm.MatchFields = []v1.NodeSelectorRequirement{nodeSelReq}
+		} else {
+			keyNodeNameFounded := false
```
ahmed-mez, line 65: nit
```suggestion
			keyNodeNameFound := false
```

pkg/controller/utils/affinity/affinity.go
```diff
@@ -55,11 +55,27 @@ func ReplaceNodeNameNodeAffinity(affinity *v1.Affinity, nodename string) *v1.Aff
 	}
 
 	// Replace node selector with the new one.
```
ahmed-mez, line 57: we should update this comment

ahmed-mez (approved):

PR #43, opened 2020-10-01
Allow the EDS to deploy pods on unschedulable nodes

L3n41c:
### What does this PR do?

Allow the `ExtendedDaemonSet` controller to deploy pods on unschedulable nodes.

### Motivation

Have the same behaviour as with the regular Kubernetes `DaemonSet`.
https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#taints-and-tolerations

### Additional Notes

### Describe your test plan

Cordon a node and check that the EDS controller still deploy a pod on it.


vboulineau:
We should be able to test this in E2E tests by building Nodes with custom taints.

vboulineau (approved):

PR #50, opened 2020-10-09
Add eds_controller_leader gauge to identify leader from /metrics and /ksmetrics

vboulineau:
### What does this PR do?

Add eds_controller_leader gauge to identify leader from `/metrics` and `/ksmetrics`

### Motivation

### Additional Notes

### Describe your test plan

Spawn two replicas of EDS instance, run 
`k exec -ti $POD -- curl http://localhost:8080/metrics`
`k exec -ti $POD -- curl http://localhost:8080/ksmetrics`

You should see `eds_controller_leader` set to 1 with `leader: true` or `leader: false` if leader or follower.

### DIFF
--- a/pkg/controller/metrics/leader.go
+++ b/pkg/controller/metrics/leader.go
@@ -1,0 +1,42 @@
+package metrics
+
+import (
+	"strconv"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+const (
+	leaderLabel = "leader"
+)
+
+var (
+	isLeader  = false
+	edsLeader = prometheus.NewGaugeVec(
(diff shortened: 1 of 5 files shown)


pkg/controller/metrics/leader.go
```diff
@@ -0,0 +1,30 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	edsLeader = prometheus.NewGauge(
```
ahmed-mez, line 9: we need labels in case we want to apply label joins

ahmed-mez (approved):
LGTM ✅ 
Please add to the description that we have new metric in `/ksmetrics` as well

PR #39, opened 2020-09-04
First version operator-sdk 1.0.0

vboulineau:

PR #51, opened 2020-10-20
add parameter to configure max restarts for canary pause

celenechang:
### What does this PR do?

Currently the Canary autopause feature pauses the canary deployment if any of the canary pod containers restart more than 0 times. We have found that this is too sensitive, so this PR makes this a configurable parameter with a default value of 2. The `maxRestarts` parameter means that this is the number of restarts of a single container that is considered tolerable before we autopause.

It also adds the ability to disable the feature entirely via the spec.

### Motivation

Better user experience

### Additional Notes

A subsequent PR with a new parameter to define a stability (no restarts) duration requirement is planned.

### Describe your test plan

Use an image such as `bash:latest` in the EDS configuration that causes the container to enter CLB. Increase the canary duration to be sufficiently large to give time for restarts.

Check if canary is automatically paused when the container restarts 3 times (default) without adding the `autoPause.maxRestarts` setting to the EDS spec.

Then try setting `autoPause.maxRestarts` to a larger number and see if the canary pauses after the corresponding number of restarts (see examples/foo-eds_v2.yaml for an example configuration).

Lastly, try setting `autoPause.enabled` to `false` and `autoPause.maxRestarts` to `0` and make sure that a restart does not pause the canary.


pkg/controller/utils/pod/pod.go
```diff
@@ -101,12 +101,13 @@ func GetPodConditionFromList(conditions []v1.PodCondition, conditionType v1.PodC
 
 // IsPodRestarting checks if a pod in the Canary deployment is restarting
 // This returns the "reason" for the pod with the most restarts
-func IsPodRestarting(pod *v1.Pod) (bool, datadoghqv1alpha1.ExtendedDaemonSetStatusReason) {
-	var maxRestartCount int
+func IsPodRestarting(pod *v1.Pod, maxRestartCount int) (bool, datadoghqv1alpha1.ExtendedDaemonSetStatusReason) {
+	// track the highest number of restarts among pod containers
+	var mostRestartCount int
```
ahmed-mez, line 106: what if the eds controller restarts during a canary? I think we should store the count in the CR status

api/v1alpha1/extendeddaemonset_types.go
```diff
@@ -75,6 +75,7 @@ type ExtendedDaemonSetSpecStrategyRollingUpdate struct {
 type ExtendedDaemonSetSpecStrategyCanary struct {
 	Replicas     *intstr.IntOrString   `json:"replicas,omitempty"`
 	Duration     *metav1.Duration      `json:"duration,omitempty"`
+	MaxRestarts  *intstr.IntOrString   `json:"maxRestarts,omitempty"`
```
clamoriniere, line 78: I'm wondering if it is necessary to have a `*intstr.IntOrString `. I feel like `*int` should be enough.

pkg/controller/utils/pod/pod.go
clamoriniere: `mostRestartCount` is calculated by looking at each run  the `pod.Status`, so I think we are safe if the controller restart: the function is completely stateless.

clamoriniere (commented):
Look good.

Could you add also a documentation about how configure and use the `pause` feature: 
* configuration example
* kubectl plugin command example

controllers/extendeddaemonsetreplicaset/strategy/canary.go
```diff
@@ -29,6 +36,24 @@ func ManageCanaryDeployment(client client.Client, daemonset *v1alpha1.ExtendedDa
...
+			autoPauseEnabled = defaultAutoPauseEnabled
+		} else {
+			autoPauseEnabled = *daemonset.Spec.Strategy.Canary.AutoPause.Enabled
+		}
+		if daemonset.Spec.Strategy.Canary.AutoPause.MaxRestarts == nil {
+			maxRestarts = defaultMaxRestarts
+		} else {
+			maxRestarts = int(*daemonset.Spec.Strategy.Canary.AutoPause.MaxRestarts)
+		}
+	}
```
ahmed-mez, line 55: please correct me if I'm wrong, but I think at this point the defaulting should be already done via `DefaultExtendedDaemonSetSpecStrategyCanaryAutoPause` ?

ahmed-mez: to be clearer, the question is: why do we default the autopause config twice?

examples/kind-cluster-configuration.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Cluster
```
ahmed-mez, line 1: nit: we should probably move this under `hack` ?  as `examples` should probably contain CR examples only

api/v1alpha1/extendeddaemonset_types.go
```diff
@@ -77,7 +77,15 @@ type ExtendedDaemonSetSpecStrategyCanary struct {
...
-	NodeAntiAffinityKeys []string `json:"nodeAntiAffinityKeys,omitempty"`
+	NodeAntiAffinityKeys []string                                      `json:"nodeAntiAffinityKeys,omitempty"`
+	AutoPause            *ExtendedDaemonSetSpecStrategyCanaryAutoPause `json:"autoPause,omitempty"`
+}
+
+// ExtendedDaemonSetSpecStrategyCanaryAutoPause defines the canary deployment AutoPause parameters of the ExtendedDaemonSet
+// +k8s:openapi-gen=true
+type ExtendedDaemonSetSpecStrategyCanaryAutoPause struct {
+	Enabled     *bool  `json:"enabled,omitempty"`
+	MaxRestarts *int32 `json:"maxRestarts,omitempty"`
```
ahmed-mez, line 88: could we please document these fields? especially `MaxRestarts`, we should make it clear that it corresponds to the max number of restarts per canary pod, not the sum of restarts of all canary pods

examples/kind-cluster-configuration.yaml
clamoriniere: I think we propose to use this file in the documentation for testing the example.
`hack` folder is more for contributor.

ahmed-mez: okay

controllers/extendeddaemonsetreplicaset/strategy/canary.go
celenechang: Thanks for bringing this up, as I didn't have a complete grasp on this.

It seems we `default` the ERS (in extendeddaemonset_default.go) if the minimum required settings are not defined. In this case we could set the default AutoPause settings with `DefaultExtendedDaemonSetSpecStrategyCanaryAutoPause()`, but it's not strictly necessary since we also have this defaulting code here in canary.go. If you think it's confusing to include the change in extendeddaemonset_default.go I can remove it.

controllers/extendeddaemonset/controller.go
```diff
@@ -262,9 +262,11 @@ func (r *Reconciler) updateInstanceWithCurrentRS(logger logr.Logger, daemonset *
 
 	// Check if newDaemonset differs from existing daemonset, and update if so
 	if !apiequality.Semantic.DeepEqual(daemonset, newDaemonset) {
+
 		if updateDaemonsetSpec {
-			if err := r.client.Update(context.TODO(), newDaemonset); err != nil {
-				return newDaemonset, reconcile.Result{}, err
+			newDaemonsetCopy := newDaemonset.DeepCopy()
```
celenechang, line 267: I'm not sure why this DeepCopy() call seems to have been removed in #39 . Without it both the spec and the status do not update properly; only the one called first gets updated.

controllers/extendeddaemonsetreplicaset/strategy/canary.go
ahmed-mez: thanks for clarifying this
tbh I would expect the defaulting to be done in one place (in `extendeddaemonset_default.go`), the fact that the same defaulting logic and the same constants (`defaultMaxRestarts` `defaultAutoPauseEnabled` `defaultCanaryAutoPauseEnabled` `defaultCanaryAutoPauseMaxRestarts`) are defined in different packages doesn't seem right 🤔

clamoriniere: I think if we do it as @ahmed-mez  said in `extendeddaemonset_default.go`, the defaulting is not needed in `ManageCanaryDeployment` because we always call the default method.
To achieve it, the function  `IsDefaultedExtendedDaemonSet()` [link](https://github.com/DataDog/extendeddaemonset/blob/master/api/v1alpha1/extendeddaemonset_default.go#L25), should be updated, to return false if the canary configuration is missing default values.

celenechang: I tested it like this, with only setting the defaults in extendeddaemonset_default.go, and it led to undesired behaviors. I will try again.

celenechang: OK got this fixed, and confirmed it works whether `autoPause`, `autoPause.Enabled`, or `autoPause.MaxRestarts` is missing. Sorry for the confusion

controllers/extendeddaemonset/controller.go
celenechang: this is now resolved

README.md
```diff
@@ -241,12 +265,44 @@ Usage:
...
+
+The Canary deployment can be manually failed. This command will restore the currently active ExtendedReplicaSet on the Canary pods.
+
+`kubectl-eds canary fail <ExtendedDaemonSet name>`
+
+#### Reset Canary deployment
+
+Following failure of the Canary deployment, the `fail` annotation should be reset with this command.
+
+`kubectl-eds canary reset <ExtendedDaemonSet name>`
```
ahmed-mez, line 302: thanks that's super useful

controllers/extendeddaemonsetreplicaset/strategy/canary.go
```diff
@@ -17,6 +17,13 @@ import (
 	podUtils "github.com/DataDog/extendeddaemonset/pkg/controller/utils/pod"
 )
 
+// const (
+// 	// defaultAutoPauseEnabled enables the AutoPause feature if the spec field is not defined
+// 	defaultAutoPauseEnabled = true
+// 	// defaultMaxRestarts defines the default threshold of tolerable restart counts beyond which the canary deployment should pause
+// 	defaultMaxRestarts = 2
+// )
```
ahmed-mez, line 25: ✂️

pkg/controller/utils/pod/pod.go
```diff
@@ -101,12 +101,13 @@ func GetPodConditionFromList(conditions []v1.PodCondition, conditionType v1.PodC
 
 // IsPodRestarting checks if a pod in the Canary deployment is restarting
 // This returns the "reason" for the pod with the most restarts
-func IsPodRestarting(pod *v1.Pod) (bool, datadoghqv1alpha1.ExtendedDaemonSetStatusReason) {
-	var maxRestartCount int
+func IsPodRestarting(pod *v1.Pod, maxRestartCount int) (bool, datadoghqv1alpha1.ExtendedDaemonSetStatusReason) {
```
ahmed-mez, line 104: I know this function wasn't introduced in this PR but it would be great to unit test it

ahmed-mez (commented):
looks great! just few comments

ahmed-mez (approved):

PR #56, opened 2020-11-02
Add canary paused / failed metrics

xornivore:
### What does this PR do?

Adds a new `eds_status_canary_paused` gauge metric to identify canary in paused state along with the reason.
Adds a new `eds_status_canary_failed` gauge metric to identify canary in failed state along with the reason.

### Motivation

Canary metrics

### Describe your test plan

Check the metrics are exported correctly for canaries in active and paused state.

PR #60, opened 2020-11-05
fix canary fail metric

celenechang:
### What does this PR do?

Follow up to https://github.com/DataDog/extendeddaemonset/pull/56

Since we set `Status.Canary` to `nil` when the canary deployment fails, this metric will never change.

### Describe your test plan

Deploy a canary EDS, and then `exec` onto the EDS controller pod and check the fail metric from the output of `curl localhost:8080/ksmetrics` . Fail the canary and check the metric again, and make sure it changed to `1`.


xornivore (approved):
🙇

ahmed-mez (changes requested):
fixes must target the v0.4 branch

PR #66, opened 2020-11-20
Add ERS canary NoRestartDuration and AutoFail configuration options to complement AutoPause behavior

xornivore:
### What does this PR do?

Adds new features to handle restarts in the canary phase:
- `NoRestartsDuration` for canary preventing end of the canary phase until there are no restarts within this time. 
- `AutoFail` behaving similarly to pause but causing canary to be failed either on `MaxRestarts` or on exceeding `MaxRestartsDuration` for pod restarts.

### Additional Notes

- `Status.Conditions` are used to record the last restart of an ERS pod and progressively track the difference between initial transition and update times.

- A future TODO - adding more pod termination statuses recorded as reason for canary pause or fail statuses, currently only `CrashLoopBackOff` and `OOMKilled` are captured, the rest is mapped to `Unknown`.

- There's quite a bit of refactoring of existing canary management logic to make it more testable with unit tests. Hoping reviewers can critique the changes and advise on possible improvements.

### Describe your test plan

For `NoRestartsDuration` - configure a canary with a pod that is causing restarts within no restart duration, make sure that canary phase is not ended until this duration elapses.

For `AutoFail` - configure it with either `MaxRestarts` or `MaxRestartsDuration` and make sure the transition to `canary-failed` state takes place.


api/v1alpha1/const.go
```diff
@@ -24,6 +24,8 @@ const (
 	ExtendedDaemonSetCanaryPausedReasonAnnotationKey = "extendeddaemonset.datadoghq.com/canary-paused-reason"
 	// ExtendedDaemonSetCanaryFailedAnnotationKey annotation key used on ExtendedDaemonset in order to detect if a canary deployment has failed.
 	ExtendedDaemonSetCanaryFailedAnnotationKey = "extendeddaemonset.datadoghq.com/canary-failed"
+	// ExtendedDaemonSetCanaryFailedReasonAnnotationKey annotation key used on ExtendedDaemonset in order to provide a reason that the a canary deployment is failed.
+	ExtendedDaemonSetCanaryFailedReasonAnnotationKey = "extendeddaemonset.datadoghq.com/canary-failed-reason"
```
clamoriniere, line 28: should we put the reason in the status instead?

xornivore: We still need to keep these annotations because they are at the EDS level and if something goes wrong with the ERS we won't be able to tell, besides we have a command to fail the canary via the annotation. For now I added `canary-failed` status of the canary to keep track of it in the status, but it's a transient state as it will be deleted shortly after that.

controllers/extendeddaemonsetreplicaset/strategy/canary.go
```diff
@@ -1,98 +1,275 @@
...
+	// Update result to reflect active pods currently experiencing restarts
+	// potentially placing canary into paused or failed state
+	manageCanaryPodRestarts(podsToCheckForRestarts, params, result)
+
+	// Update pod counts
 	result.NewStatus.Desired = desiredPods
 	result.NewStatus.Ready = readyPods
 	result.NewStatus.Available = availablePods
 	result.NewStatus.Current = currentPods
+
```
clamoriniere, line 127: I think we don't update the `result.NewStatus.Reason` with the info: `Paused`, `Failed`

clamoriniere (commented):
really nice PR 👍 
I just added one comment

controllers/extendeddaemonsetreplicaset/strategy/canary.go
xornivore: So while there is currently no `result.NewStatus.Reason` there is `result.NewStatus.Status` which is updated to:
- `canary` for active and paused canaries.
- `canary-failed` for failed canaries.

```
type ExtendedDaemonSetReplicaSetStatus struct {
	Status                   string `json:"status"`
	Desired                  int32  `json:"desired"`
	Current                  int32  `json:"current"`
	Ready                    int32  `json:"ready"`
	Available                int32  `json:"available"`
	IgnoredUnresponsiveNodes int32  `json:"ignoredUnresponsiveNodes"`
	// Conditions Represents the latest available observations of a DaemonSet's current state.
	// +listType=map
	// +listMapKey=type
	Conditions []ExtendedDaemonSetReplicaSetCondition `json:"conditions,omitempty"`
}
```

We interpret status as an enumerable field today (as opposed to a freeform description). Example where it is being taken into consideration as such: https://github.com/DataDog/extendeddaemonset/blob/master/controllers/extendeddaemonsetreplicaset/controller.go#L183

clamoriniere (approved):

PR #67, opened 2020-12-02
[e2e] Resurrecting e2e tests on kind with envtest framework

xornivore:
### What does this PR do?

- Added initial tests for testing autoPause and autoFail of canary on pod restarts.

### Motivation

- While envtest control plane tests are great they don't allow us to exercise scenarios that assert on behaviors involving failure cases such as pod restarts.

### Additional Notes

### Describe your test plan

### DIFF
--- a/controllers/extendeddaemonset_e2e_test.go
+++ b/controllers/extendeddaemonset_e2e_test.go
@@ -3,7 +3,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-2019 Datadog, Inc.
 
-// +build !e2e
+// +build e2e
 
 package controllers
 
@@ -13,15 +13,12 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/google/go-cmp/cmp"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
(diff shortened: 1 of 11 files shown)


controllers/suite_e2e_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+func TestAPIs(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecsWithDefaultAndCustomReporters(t,
+		"Controller Suite",
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = BeforeSuite(func(done Done) {
```
vboulineau, line 53: Are we able to have that in common with the normal tests?

controllers/extendeddaemonset_e2e_test.go
```diff
@@ -0,0 +1,311 @@
...
+	const (
+		timeout     = 1 * time.Minute
+		longTimeout = 5 * time.Minute
+		interval    = 2 * time.Second
+	)
+
+	intString2 := intstr.FromInt(2)
+	intString10 := intstr.FromInt(10)
+
+	Context("Initial deployment", func() {
```
vboulineau, line 41: I think we should have specific, dedicated tests if we are to introduce E2E with Cluster.

controllers/suite_e2e_test.go
xornivore: We can extract some parts here but there are differences:
- CRD versions (v1 does not install on k8s 1.19.1
- Use existing cluster boolean
- Node creation
I will try to deduplicate this a bit.

controllers/extendeddaemonset_e2e_test.go
xornivore: Do you mean splitting this monolithic one into separate tests? There is still a need for the initial deployment/canary phase before you can do anything else. Concern is duration of tests as well if we repeat that on every split test.

controllers/suite_e2e_test.go
vboulineau: V1 not installing is definitely an issue we should fix, it's supposed to work in 1.16+.
For the other two we should have some parameters or functors to customize setup

controllers/extendeddaemonset_e2e_test.go
vboulineau: I meant the duplication of tests, I believe these tests are duplicated from the original file:

```
It("Should handle EDS"
It("Should do canary deployment"
It("Should add canary labels"
It("Should remove canary labels"
```

xornivore: Yes, so we can just wait for canary to be over. I will remove these.

clamoriniere (approved):
Awesome work @xornivore 🥇

controllers/extendeddaemonset_test.go
```diff
@@ -57,145 +61,81 @@ var _ = Describe("ExtendedDaemonSet Controller", func() {
...
 				return len(canaryPods.Items) == 0
-			}, timeout, interval).Should(BeTrue())
+			}), timeout, interval).Should(BeTrue())
 		})
 	})
 
 	Context("Using ExtendedDaemonsetSetting", func() {
-		var err error
-		namespace := "default"
+		// var err error
```
clamoriniere, line 138: cleanup 
```suggestion
```

PR #68, opened 2020-12-04
[eds] Introduce PodCannotStart condition and canary auto-pause handling

xornivore:
### What does this PR do?

- Introduces logic to handle certain failure waiting states as cause for auto-pausing canary.
- Simplifies logic for reported container states in terminated reason (will basically take anything reported from container).
- e2e tests for `PodCannotStart` conditions/scenarios (container config error and image pull errors).

### Motivation

Current limitations in support for container waiting state which are abnormal and should result in pausing canary.

### Additional Notes

### Describe your test plan

Scenarios similar to tests in e2e.


controllers/extendeddaemonsetreplicaset/strategy/canary.go
```diff
@@ -164,20 +163,27 @@ func manageCanaryPodRestarts(pods []*v1.Pod, params *Parameters, result *Result)
...
+			}
 		}
 
-		restartTime, recentRestartReason := podUtils.MostRecentRestart(pod)
-		if restartTime.After(newRestartTime) {
-			newRestartTime = restartTime
-			restartingPodStatus = fmt.Sprintf("Pod %s restarting with reason: %s", pod.ObjectMeta.Name, string(recentRestartReason))
+		var cannotStartReason v1alpha1.ExtendedDaemonSetStatusReason
+		cannotStart, cannotStartReason = podUtils.CannotStart(pod)
+		if cannotStart {
```
clamoriniere, line 185: could you add a counter about how many pods are note able to start? and add this counter to the description?

xornivore: We could do that but I think we should refrain from converting EDS status into a fully fledged debugging tool. Rationale for the currently implemented behavior is that we are considering it to be abnormal enough to pause the canary on first failure.

pkg/controller/utils/pod/pod.go
```diff
@@ -130,22 +130,37 @@ func MostRecentRestart(pod *v1.Pod) (time.Time, datadoghqv1alpha1.ExtendedDaemon
...
-		return datadoghqv1alpha1.ExtendedDaemonSetStatusReasonCLB
-	case string(datadoghqv1alpha1.ExtendedDaemonSetStatusReasonOOM):
-		return datadoghqv1alpha1.ExtendedDaemonSetStatusReasonOOM
-	default:
-		return datadoghqv1alpha1.ExtendedDaemonSetStatusReasonUnknown
+var cannotStartReasons = []string{
+	"ErrImagePull",
+	"ImagePullBackOff",
+	"CreateContainerConfigError",
+}
```
clamoriniere, line 144: could you add other reasons defined here: https://github.com/kubernetes/kubernetes/blob/88512be213159f87ed2dad07f75c02d355eadac5/pkg/kubelet/kuberuntime/kuberuntime_container.go#L58

examples/foo-eds_v4_image_pull_error.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: datadoghq.com/v1alpha1
```
clamoriniere, line 1: I think we should not put e2e tests manifest in another than the `examples` folder. maybe `controllers/test-data`

xornivore: What about if we add a `README.md` under examples - that way we can explain what each example intends to demonstrate. The e2e tests don't depend on these new files per se, my thought was to demonstrate certain features and behaviors per se, but I agree that it might be confusing to others.

Or maybe you meant that these should be only showing settings and in the happy case only. These indeed will cause canary to handle in abnormal scenarios.

clamoriniere: Yes I mean `examples` should only contains working examples.
But it could be also interesting for users, to see how the canary works. So maybe you can add a subfolder `examples/canary-tests`. WDYT ?

clamoriniere (approved):

PR #69, opened 2020-12-21
use client.Patch() for [add|remov]ing canary label

clamoriniere:
### What does this PR do?

Use the `client.Patch()` function instead of `client.Update()` to add canary label on Pods.

### Motivation

Previously the function client.Update() was used to add or remove the canary
labels on the Pods. Using `Patch()` limit possible conflict and decrease message
size.

It removes bug on Kubernetes 1.19.0 when the API-Server return an error with the function `client.Update()`, because the controller is trying to update an immutable `Pod.Spec` field.  

### Additional Notes

### Describe your test plan

* Create a Kubernetes v1.19
* Deploy a EDS with a canary configuration
* Update the EDS.spec.template
* check that canary Pods have the canary labels.


xornivore (approved):
Do you want to do the same for EDS annotations while you are at it? 😄

PR #71, opened 2020-12-24
[Fix] Make sure the ERS only handles defaulted EDS CR

ahmed-mez:
### What does this PR do?

- Make the ERS controller requeue the request if the CR is not defaulted (yet) by the EDS controller

### Motivation

By design, the ERS controller can receive requests before the EDS controller defaults the CR.

This PR adds safeguard against nil pointers when accessing CR fields (e.g `Spec.Strategy.ReconcileFrequency.Duration`) in the ERS controller.

### Additional Notes

Related to https://github.com/DataDog/extendeddaemonset/pull/65

### Describe your test plan

Same as https://github.com/DataDog/extendeddaemonset/pull/65

- Create a new ExtendedDaemonset. The reconcileFrequency should have been defaulted when
the CR is created.
- Edit this ExtendedDaemonset and remove the reconcileFrequency from the spec.strategy;
The controller should update the spec.strategy to add it back.
- Another test is to set a different reconcileFrequency value, to validate that the controller
Didn't override a value provided by the user.

### DIFF
--- a/api/v1alpha1/extendeddaemonset_default.go
+++ b/api/v1alpha1/extendeddaemonset_default.go
@@ -32,10 +32,6 @@ func IsDefaultedExtendedDaemonSet(dd *ExtendedDaemonSet) bool {
 		return false
 	}
 
-	if dd.Spec.Strategy.ReconcileFrequency == nil {
-		return false
-	}
-
 	if dd.Spec.Strategy.Canary != nil {
 		if defaulted := IsDefaultedExtendedDaemonSetSpecStrategyCanary(dd.Spec.Strategy.Canary); !defaulted {
 			return false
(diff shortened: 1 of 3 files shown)


api/v1alpha1/extendeddaemonset_default.go
```diff
@@ -32,10 +32,6 @@ func IsDefaultedExtendedDaemonSet(dd *ExtendedDaemonSet) bool {
 		return false
 	}
 
-	if dd.Spec.Strategy.ReconcileFrequency == nil {
-		return false
-	}
```
ahmed-mez, line 37: This is duplicated (see below in this function)

xornivore (approved):
👍

PR #74, opened 2020-12-29
Improve controller logic + fix tests flakiness

clamoriniere:
### What does this PR do?

* Add new CanaryFailed and CanaryPause conditions in ERS.status
* Add `EDS.status.conditions` with CanaryFailed and CanaryPause
  conditions.
* Remove EDS annotations update in the ERS reconcile loop.
* Check ERS.status.conditions in EDS reconcile loop to detec Canary
  paused or failed
* Update unit-tests and e2t tests according the logic update
* Add some reconcile loops documentations.
* In the ERS-controller, If the parent `ExtendedDaemonset` is not defaulted, we force the Defaulting by updating the ERS status with an error conditions. like this the EDS-controller received an event on an ERS update and requeue the parent EDS.
* Speed-up reconcile frequency in tests to avoid test flakiness
* Use only one Canary Pod in the canary strategy configuration in test.

### Motivation

Remove test flakiness and improve EDS/ERS lifecycles.

### Additional Notes

N/A

### Describe your test plan

* Tests should not be flaky.


controllers/extendeddaemonset/controller_test.go
```diff
@@ -776,11 +777,11 @@ func TestReconcileExtendedDaemonSet_updateInstanceWithCurrentRS(t *testing.T) {
 				t.Errorf("ReconcileExtendedDaemonSet.updateInstanceWithCurrentRS() error = %v, wantErr %v", err, tt.wantErr)
 				return
 			}
-			if !apiequality.Semantic.DeepEqual(got, tt.want) {
-				t.Errorf("ReconcileExtendedDaemonSet.updateInstanceWithCurrentRS() got = %#v, \n want %#v", got, tt.want)
+			if diff := cmp.Diff(tt.want, got); diff != "" {
```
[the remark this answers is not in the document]
clamoriniere: done

controllers/extendeddaemonsetreplicaset/strategy/utils.go
```diff
@@ -178,6 +183,7 @@ func failCanaryDeployment(client client.Client, eds *datadoghqv1alpha1.ExtendedD
 	)
 }
 
+/*
 func refetchPod(c client.Client, pod *corev1.Pod) (*corev1.Pod, error) {
```
L3n41c, line 187: Could be completely deleted.

controllers/extendeddaemonsetreplicaset/strategy/utils.go
```diff
@@ -187,51 +193,58 @@ func refetchPod(c client.Client, pod *corev1.Pod) (*corev1.Pod, error) {
...
 	}
 
-	pod, err := refetchPod(c, pod)
-	if err != nil {
-		return err
-	}
+	//pod, err := refetchPod(c, pod)
+	//if err != nil {
+	//	return err
+	//}
```
L3n41c, line 207: Could be completely deleted.
```suggestion
```

controllers/extendeddaemonsetreplicaset/strategy/utils.go
```diff
@@ -187,51 +193,58 @@ func refetchPod(c client.Client, pod *corev1.Pod) (*corev1.Pod, error) {
...
```
L3n41c, line 235: Could be completely deleted.
```suggestion
```

controllers/extendeddaemonset/controller.go
```diff
@@ -510,6 +519,13 @@ func (r *Reconciler) cleanupReplicaSet(logger logr.Logger, rsList *datadoghqv1al
 	return utilserrors.NewAggregate(errs)
 }
 
+func clearFailureAnnotations(eds *datadoghqv1alpha1.ExtendedDaemonSet) {
```
celenechang, line 522: i think in this case we no longer need https://github.com/DataDog/extendeddaemonset/blob/master/pkg/plugin/canary/fail.go#L85

clamoriniere: Yes indeed.

celenechang: do you think it is useful to maintain a `FailedReason` annotation? it is unlikely to be seen

controllers/extendeddaemonset/controller.go
```diff
@@ -256,44 +284,32 @@ func (r *Reconciler) updateInstanceWithCurrentRS(logger logr.Logger, daemonset *
...
-			newDaemonsetCopy.Status = newDaemonset.Status
-			logger.Info("Updating ExtendedDaemonSet status")
-			if err := r.client.Status().Update(context.TODO(), newDaemonsetCopy); err != nil {
-				logger.Error(err, "Failed to update ExtendedDaemonSet status")
-				return newDaemonsetCopy, reconcile.Result{}, err
-			}
-			return newDaemonsetCopy, reconcile.Result{}, nil
+		// Updating the status in any case.
+		// Make and use a copy because undesired behaviors occur when making two update calls
+		// The returned EDS instance has losted the ExtendedDaemonSet.spec update part.
```
celenechang, line 296: suggestion: `The EDS instance will lose its new spec as a result of the Status update`

controllers/extendeddaemonset/controller.go
clamoriniere: I wanted to maintain "Failed" annotation to allow a user to failed manually a EDS canary

clamoriniere: the controller only remove annotations (paused and failed) for the auto-failed usecase.
I did it like this, because by updating the EDS.spec with the "active" ERS, it removes (delete) the canary ERS. So for me the pause and failed annotation doesn't make sense anymore. 

About the `FailedReason` we can renamed it `StatusReason`?

celenechang: apologies, i clarified my question above regarding a more general name for the function, rather than the annotation.

i understand, once the spec is updated the annotations will be stale. do you mind clarifying: 

> the controller only remove annotations (paused and failed) for the auto-failed usecase

how are we distinguishing between autofail and manual fail?

pkg/plugin/canary/fail.go
```diff
@@ -81,31 +80,6 @@ func newCmdFail(streams genericclioptions.IOStreams) *cobra.Command {
 	return cmd
 }
 
-// newCmdReset provides a cobra command wrapping failOptions
```
celenechang, line 84: thanks! sorry to be a pain, this should be removed as well https://github.com/DataDog/extendeddaemonset/blob/87217d683f4148d582037db50b579f0796ac881f/README.md#reset-canary-deployment

controllers/extendeddaemonset/controller.go
clamoriniere: * auto-fail is detected/trigged in the EDS-reconcile loop thanks to a CanaryFailed condition the `ERS.status.conditions`
* manual fail is detected/trigged thanks to the CanaryFailed annotation added by the user on the EDS instance.

clamoriniere: about your initial comments, wdyt about `clearFailureAnnotations`  -> `clearCanaryAnnotations` ?

celenechang: i see; but in either case, `IsCanaryFailed` is `true`, and this seems to be the only requirement to call `clearFailureAnnotations`

clamoriniere: clearFailedAnnotations are call only in the `case IsCanaryFailed`, and it correspond only to the CanaryFailed condition or annotation

for the CanaryPause condition and annotation the controller have another boolean: isCanaryPaused

pkg/plugin/canary/fail.go
clamoriniere: no problem :) 
PR updated

controllers/extendeddaemonset/controller.go
celenechang: oh sorry, i misunderstood what you meant. i read 

> the controller only remove annotations (paused and failed) for the auto-failed usecase

as meaning that the manual fail case would not have the annotations removed.

that brings me back to my original question though, where I think a `FailedReason` annotation will never be observed (since it will be cleared on the first reconcile loop following the failure), unless we persist in another way (e.g. a metric tag)?

controllers/extendeddaemonset/controller.go
```diff
@@ -256,44 +284,32 @@ func (r *Reconciler) updateInstanceWithCurrentRS(logger logr.Logger, daemonset *
...
-			logger.Info("Updating ExtendedDaemonSet status")
-			if err := r.client.Status().Update(context.TODO(), newDaemonsetCopy); err != nil {
-				logger.Error(err, "Failed to update ExtendedDaemonSet status")
-				return newDaemonsetCopy, reconcile.Result{}, err
-			}
-			return newDaemonsetCopy, reconcile.Result{}, nil
+		// Updating the status in any case.
+		// Make and use a copy because undesired behaviors occur when making two update calls
+		// The EDS instance will lose its new spec as a result of the Status update
+		extendedDaemonsetCopy := newDaemonset.DeepCopy()
```
xornivore, line 297: Having to do a `DeepCopy` is a bit odd, can we not do a `Patch` instead?

xornivore (commented):
Great changes, and much cleaner now. 💯 

Mind checking if we have this e2e test (I suspect we don't):
working eds  >> update eds >> (active canary with bad change) >> update with fix while canary active >> working eds

controllers/extendeddaemonset/controller.go
clamoriniere: to make a patch we also need to do a DeepCopy() to build the diff

controllers/extendeddaemonset/controller.go
```diff
@@ -510,6 +526,13 @@ func (r *Reconciler) cleanupReplicaSet(logger logr.Logger, rsList *datadoghqv1al
 	return utilserrors.NewAggregate(errs)
 }
 
+func clearCanaryAnnotations(eds *datadoghqv1alpha1.ExtendedDaemonSet) {
+	delete(eds.Annotations, datadoghqv1alpha1.ExtendedDaemonSetCanaryPausedAnnotationKey)
```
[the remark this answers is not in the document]
clamoriniere: maybe we can. but I'm not sure it will improve the readability or decrease the number of line since we will need to create a slice of annotation's keys.

controllers/extendeddaemonset/conditions/update.go
```diff
@@ -0,0 +1,96 @@
...
+		Status:             conditionStatus,
+		LastUpdateTime:     now,
+		LastTransitionTime: now,
+		Reason:             reason,
+		Message:            message,
+	}
+}
+
+// UpdateExtendedDaemonSetStatusCondition used to update a specific ExtendedDaemonSetConditionType
+func UpdateExtendedDaemonSetStatusCondition(status *datadoghqv1alpha1.ExtendedDaemonSetStatus, now metav1.Time, t datadoghqv1alpha1.ExtendedDaemonSetConditionType, conditionStatus corev1.ConditionStatus, reason, desc string, writeFalseIfNotExist, supportLastUpdate bool) {
```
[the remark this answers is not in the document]
clamoriniere: do you think I should create a `struct Options` for clarity?

like:

```golang
type ConditionOptions struct {
    IgnoreFalseConditionIfNotExist bool
    SupportLastUpdate              bool
}

**Comment by clamoriniere** [comment]
@xornivore 

> Mind checking if we have this e2e test (I suspect we don't):
> working eds >> update eds >> (active canary with bad change) >> update with fix while canary active >> working eds

Good idea I will try to add this test

File: controllers/extendeddaemonset/conditions/update.go (hunk 10 above)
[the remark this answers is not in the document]
Comment (xornivore) [reply controllers/extendeddaemonset/conditions/update.go:29]: That or more helpers instead of one, maybe using a struct like this. Good opportunity to document what these flags are doing.

File: controllers/extendeddaemonset/controller.go (hunk 8 above)
Comment (xornivore) [reply controllers/extendeddaemonset/controller.go:297]: Ok, is the comment up-to-date? `when making two update calls` - is this referring to making the two updates in this func here? Can we consolidate them into one?

File: controllers/extendeddaemonset/controller.go [hunk 11]
```diff
@@ -215,23 +215,47 @@ func (r *Reconciler) updateInstanceWithCurrentRS(logger logr.Logger, daemonset *
...
+		case isCanaryFailed:
 			// Canary deployment is no longer needed because it was marked as failed
 			// This `if` block needs to be first to respect Failed Canary state until annotation is removed
 			newDaemonset.Status.Canary = nil
 			newDaemonset.Status.State = datadoghqv1alpha1.ExtendedDaemonSetStatusStateCanaryFailed
+			newDaemonset.Status.Reason = ""
 			// Restore active replicaset template. Note: this requires a full daemonset update
 			newDaemonset.Spec.Template = current.Spec.Template
 			updateDaemonsetSpec = true
+			clearCanaryAnnotations(newDaemonset)
```
celenechang, line 253: i had another thought - what do you think of moving this function to the next `case` block? that way we can reset all canary annotations when a canary ends. this came up while adding additional functionality to the `unpause` feature and needed to reset an annotation that indicates a canary was manually unpaused.

clamoriniere: the problem is that we need to remove the annotations when we detect the canary failed. 
if we don't remove the annotations in the `case` we will always enter to this case.

But I'm agree with you, we can also call this `clearCanaryAnnotations ` in the next `case`.
I'm doing the change

controllers/extendeddaemonset/controller.go
```diff
@@ -215,35 +215,29 @@ func (r *Reconciler) updateInstanceWithCurrentRS(logger logr.Logger, daemonset *
...
-			}
-			newDaemonset.Status.Desired += upToDate.Status.Desired
-			newDaemonset.Status.UpToDate += upToDate.Status.Available
-			newDaemonset.Status.Available += upToDate.Status.Available
-			newDaemonset.Status.IgnoredUnresponsiveNodes += upToDate.Status.IgnoredUnresponsiveNodes
-
-			newDaemonset.Status.Canary.ReplicaSet = upToDate.Name
+		}
 
+		switch isCanaryActive {
```
xornivore, line 238: can we do `if`/`else` instead here?

xornivore (approved):
Looks great, the diagram is very helpful.

Are we confident we don't need `reset` command anymore?

controllers/extendeddaemonset/controller.go
celenechang: ah, right. OK thank you!

controllers/extendeddaemonset/controller.go
```diff
@@ -425,6 +411,77 @@ func (r *Reconciler) selectNodes(logger logr.Logger, daemonsetSpec *datadoghqv1a
 	return nil
 }
 
+func isCanaryActive(daemonset *datadoghqv1alpha1.ExtendedDaemonSet, activeERSName string, uptodateERSName string, isCanaryFailed bool) bool {
```
celenechang, line 414: nit: `uptodateERSName` -> `upToDateERSName`

controllers/extendeddaemonset/controller.go
```diff
@@ -425,6 +411,77 @@ func (r *Reconciler) selectNodes(logger logr.Logger, daemonsetSpec *datadoghqv1a
...
+}
+
+func manageCanaryStatus(status *datadoghqv1alpha1.ExtendedDaemonSetStatus, upToDate *datadoghqv1alpha1.ExtendedDaemonSetReplicaSet, isCanaryActive bool, isCanaryFailed bool, isCanaryPaused bool) *datadoghqv1alpha1.ExtendedDaemonSetStatus {
+	switch {
+	case isCanaryFailed:
+		// Canary deployment is no longer needed because it was marked as failed
+		// This `if` block needs to be first to respect Failed Canary state until annotation is removed
+		status.Canary = nil
+		status.State = datadoghqv1alpha1.ExtendedDaemonSetStatusStateCanaryFailed
+		status.Reason = ""
```
celenechang, line 457: do you think it's worth it to put the failed reason here?

controllers/extendeddaemonset/controller.go
```diff
@@ -425,6 +411,77 @@ func (r *Reconciler) selectNodes(logger logr.Logger, daemonsetSpec *datadoghqv1a
...
+
+func manageCanaryStatusConditions(status *datadoghqv1alpha1.ExtendedDaemonSetStatus, now metav1.Time, isCanaryFailed bool, isCanaryPaused bool, pausedReason datadoghqv1alpha1.ExtendedDaemonSetStatusReason, ersName string) *datadoghqv1alpha1.ExtendedDaemonSetStatus {
+	updateOptions := &conditions.UpdateConditionOptions{
+		IgnoreFalseConditionIfNotExist: false,
+		SupportLastUpdate:              false,
+	}
+	if isCanaryFailed {
+		msg := fmt.Sprintf("canary failed with ers: %s", ersName)
+		conditions.UpdateExtendedDaemonSetStatusCondition(status, now, datadoghqv1alpha1.ConditionTypeEDSCanaryFailed, corev1.ConditionTrue, "CanaryFailed", msg, updateOptions)
+		status.State = datadoghqv1alpha1.ExtendedDaemonSetStatusStateCanaryFailed
```
celenechang, line 434: i think this is being set here and in `manageCanaryStatus`

controllers/extendeddaemonset/controller.go
```diff
@@ -425,6 +411,77 @@ func (r *Reconciler) selectNodes(logger logr.Logger, daemonsetSpec *datadoghqv1a
...
+		}
+		status.Desired += upToDate.Status.Desired
+		status.UpToDate += upToDate.Status.Available
+		status.Available += upToDate.Status.Available
+		status.IgnoredUnresponsiveNodes += upToDate.Status.IgnoredUnresponsiveNodes
+
+		if !isCanaryPaused {
+			status.State = datadoghqv1alpha1.ExtendedDaemonSetStatusStateCanary
+		} else {
+			status.State = datadoghqv1alpha1.ExtendedDaemonSetStatusStateCanaryPaused
```
celenechang, line 471: similarly maybe `pausedReason` should be set here

controllers/extendeddaemonset/controller.go
clamoriniere: sure

controllers/extendeddaemonset/controller.go
clamoriniere: no really because it will be cleanup right after, when the controller will consider currentERS as active in the next reconcile loop.

But, to know the fail reason, we can now look at the failed condition that will not be removed. (only step to false, we the EDS will be back to currentERS)

controllers/extendeddaemonset/controller.go
clamoriniere: right removing this line

controllers/extendeddaemonset/controller.go
clamoriniere: good catch 👍 
done

controllers/extendeddaemonset/controller.go
```diff
@@ -425,6 +411,78 @@ func (r *Reconciler) selectNodes(logger logr.Logger, daemonsetSpec *datadoghqv1a
...
+		conditions.UpdateExtendedDaemonSetStatusCondition(status, now, datadoghqv1alpha1.ConditionTypeEDSCanaryFailed, corev1.ConditionTrue, "CanaryFailed", msg, updateOptions)
+	} else {
+		conditions.UpdateExtendedDaemonSetStatusCondition(status, now, datadoghqv1alpha1.ConditionTypeEDSCanaryFailed, corev1.ConditionFalse, "", "", updateOptions)
+	}
+
+	if isCanaryPaused && !isCanaryFailed {
+		msg := fmt.Sprintf("canary paused with ers: %s", ersName)
+		conditions.UpdateExtendedDaemonSetStatusCondition(status, now, datadoghqv1alpha1.ConditionTypeEDSCanaryPaused, corev1.ConditionTrue, string(pausedReason), msg, updateOptions)
+		status.State = datadoghqv1alpha1.ExtendedDaemonSetStatusStateCanaryPaused
+		status.Reason = pausedReason
```
celenechang, line 442: these two should be removed also

PR #77, opened 2021-01-20
persist manual unpause until end of a canary deployment, and then reset

celenechang:
### What does this PR do?

Previously, the ability to manually `unpause` a Canary deployment was added. However, if `AutoPause` is enabled and the Canary deployment exceeds the number of `maxRestarts` (for instance), then the Canary will be automatically paused again following an `unpause` action.

This PR persists a manual `unpause` action by storing it in an annotation, and inhibits the `AutoPause` feature until the Canary deployment ends and the annotation is removed.

### Motivation

Better user experience

### Additional Notes

Note that a user cannot `unpause` to effectively disable `AutoPause`, and then change their mind and re-enable `AutoPause` during the ongoing Canary deployment using the plugin. (They can manually edit the unpause annotation as a workaround.) Manual `pause`/`unpause` actions can be executed with the plugin at will.

### Describe your test plan

- `make build` (this also builds the plugin)
- Start a `kind` cluster
- Edit `config/manager/manager.yaml` and `config/manager/kustomization.yaml` to use image `test/extendeddaemonset:test` and image pull policy `Never`
- `make install`
- `make IMG=test/extendeddaemonset:test docker-build` 
- `kind load docker-image test/extendeddaemonset:test`
- `make deploy`
- Then apply the `ExtendedDaemonSet`s in `examples/` to start a canary deployment
- `bin/kubectl-eds canary pause foo`; verify it is paused (with `kubectl-eds get` and by checking the annotation)
- `bin/kubectl-eds canary unpause foo`; verify it is unpaused
- When canary deployment finishes, verify the `canary-unpaused` annotation is cleared
- Try a test with a canary deployment that is in CLB (can use `bash:latest` image) and `AutoPause` is enabled. Make sure that, after the canary is automatically paused, `bin/kubectl-eds canary unpause foo` allows the canary deployment to finish despite the restarts.
- Other behaviors to try: `canary fail foo`, `canary pause foo -> canary unpause foo`

### DIFF
--- a/controllers/extendeddaemonset/controller.go
+++ b/controllers/extendeddaemonset/controller.go
@@ -569,6 +569,7 @@ func (r *Reconciler) cleanupReplicaSet(logger logr.Logger, rsList *datadoghqv1al
 func clearCanaryAnnotations(eds *datadoghqv1alpha1.ExtendedDaemonSet) {
 	delete(eds.Annotations, datadoghqv1alpha1.ExtendedDaemonSetCanaryPausedAnnotationKey)
 	delete(eds.Annotations, datadoghqv1alpha1.ExtendedDaemonSetCanaryPausedReasonAnnotationKey)
+	delete(eds.Annotations, datadoghqv1alpha1.ExtendedDaemonSetCanaryUnpausedAnnotationKey)
 	delete(eds.Annotations, datadoghqv1alpha1.ExtendedDaemonSetCanaryFailedAnnotationKey)
 	delete(eds.Annotations, datadoghqv1alpha1.ExtendedDaemonSetCanaryFailedReasonAnnotationKey)
 }
--- a/controllers/extendeddaemonset/controller.go
+++ b/controllers/extendeddaemonset/controller.go
@@ -218,6 +218,7 @@ func (r *Reconciler) updateInstanceWithCurrentRS(logger logr.Logger, now time.Ti
 	metaNow := metav1.NewTime(now)
 
 	var updateDaemonsetSpec bool
+	var updateDaemonsetAnnotations bool
 	// If the deployment is in Canary phase, then update status (and spec as needed)
 	if daemonset.Spec.Strategy.Canary != nil {
 
@@ -252,7 +253,7 @@ func (r *Reconciler) updateInstanceWithCurrentRS(logger logr.Logger, now time.Ti
 			}
 		} else {
 			// if the Canary Deployment is not active anymore remove the canary annotations
-			clearCanaryAnnotations(newDaemonset)
+			updateDaemonsetAnnotations = clearCanaryAnnotations(newDaemonset)
 		}
 	}
 
@@ -268,10 +269,11 @@ func (r *Reconciler) updateInstanceWithCurrentRS(logger logr.Logger, now time.Ti
 			return extendedDaemonsetCopy, reconcile.Result{}, fmt.Errorf("failed to update ExtendedDaemonSet status, %w", err)
 		}
 
-		if updateDaemonsetSpec {
+		if updateDaemonsetSpec || updateDaemonsetAnnotations {
 			// we use the `extendedDaemonsetCopy` instance to have last version. that contains the latest metadata info (resource version)
 			// Copy the spec part into the extendedDaemonsetCopy.
 			extendedDaemonsetCopy.Spec = *newDaemonset.Spec.DeepCopy()
+			extendedDaemonsetCopy.Annotations = newDaemonset.Annotations
 			// In case of canaryFailed, we also update the ExtendedDaemonset.Spec
 			logger.Info("Updating ExtendedDaemonSet.Spec")
 			if err := r.client.Update(context.TODO(), extendedDaemonsetCopy); err != nil {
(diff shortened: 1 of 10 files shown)


pkg/plugin/canary/fail.go
```diff
@@ -133,16 +133,19 @@ func (o *failOptions) run() error {
...
+	if eds.Status.Canary == nil {
+		return fmt.Errorf("the ExtendedDaemonset does not have an active canary deployment")
 	}
 
 	newEds := eds.DeepCopy()
 	if newEds.Annotations == nil {
 		newEds.Annotations = make(map[string]string)
 	} else if isFailed, ok := newEds.Annotations[v1alpha1.ExtendedDaemonSetCanaryFailedAnnotationKey]; ok {
-		if o.failStatus && isFailed == "true" {
+		if o.failStatus && isFailed == "true" { //nolint:goconst
```
clamoriniere, line 146: maybe works to create a `ValueStringTrue = "true"` and `ValueStringFalse = false` in this file: https://github.com/DataDog/extendeddaemonset/blob/master/api/v1alpha1/const.go

clamoriniere: or maybe move the `stringTrue` and `stringFalse` from `pkg/plugin/canary/pause.go` to `api/v1alpha1/const.go`

clamoriniere (commented):
looks great, thank you for fixing it.
juste a small nit.

controllers/extendeddaemonset/controller.go
```diff
@@ -268,10 +269,11 @@ func (r *Reconciler) updateInstanceWithCurrentRS(logger logr.Logger, now time.Ti
 			return extendedDaemonsetCopy, reconcile.Result{}, fmt.Errorf("failed to update ExtendedDaemonSet status, %w", err)
 		}
 
-		if updateDaemonsetSpec {
+		if updateDaemonsetSpec || updateDaemonsetAnnotations {
 			// we use the `extendedDaemonsetCopy` instance to have last version. that contains the latest metadata info (resource version)
 			// Copy the spec part into the extendedDaemonsetCopy.
 			extendedDaemonsetCopy.Spec = *newDaemonset.Spec.DeepCopy()
+			extendedDaemonsetCopy.Annotations = newDaemonset.Annotations
```
celenechang, line 276: @clamoriniere this is a key change since your review; this seemed to fix an issue of annotations not being deleted (and thus the state staying in `failed`, for instance)

controllers/extendeddaemonsetreplicaset/strategy/canary.go
```diff
@@ -194,7 +196,11 @@ func manageCanaryPodFailures(pods []*v1.Pod, params *Parameters, result *Result,
 				"MaxRestarts", autoFailMaxRestarts,
 				"Reason", highRestartReason,
 			)
-		} else if !result.IsPaused && autoPauseEnabled {
+		case result.IsUnpaused:
```
celenechang, line 199: @clamoriniere this is also a key change made since your review. afaict we didn't have a way of updating the ERS `paused` condition to `false`, (so once `paused: true` we were in a positive cycle) so making it explicit here. i believe for `failed` it doesn't matter since the ERS should drop out

controllers/extendeddaemonset/controller.go
```diff
@@ -566,11 +568,22 @@ func (r *Reconciler) cleanupReplicaSet(logger logr.Logger, rsList *datadoghqv1al
...
+		datadoghqv1alpha1.ExtendedDaemonSetCanaryFailedReasonAnnotationKey,
+	}
+	var updated bool
+	for _, key := range keysToDelete {
+		if _, ok := eds.Annotations[key]; ok {
+			delete(eds.Annotations, key)
+			updated = true
+		}
+	}
+	return updated
```
clamoriniere, line 586: @xornivore will be happy about this change :)

controllers/extendeddaemonset/utils.go
```diff
@@ -60,6 +60,11 @@ func IsCanaryDeploymentEnded(specCanary *datadoghqv1alpha1.ExtendedDaemonSetSpec
 
 // IsCanaryDeploymentPaused checks if the Canary deployment has been paused
 func IsCanaryDeploymentPaused(dsAnnotations map[string]string, ers *datadoghqv1alpha1.ExtendedDaemonSetReplicaSet) (bool, datadoghqv1alpha1.ExtendedDaemonSetStatusReason) {
+	// Unpaused annotation takes precedence
+	isUnpaused, found := dsAnnotations[datadoghqv1alpha1.ExtendedDaemonSetCanaryUnpausedAnnotationKey]
+	if found && isUnpaused == v1alpha1.ValueStringTrue {
+		return false, ""
+	}
```
clamoriniere, line 67: do we still need the `IsCanaryDeploymentUnpaused()` if we consider unpause annotations here? 

or maybe:

```suggestion
	// Unpaused annotation takes precedence
	if IsCanaryDeploymentUnpaused(dsAnnotations) {
		return false, ""
	}
```

celenechang: good question, indeed we may not. will give it a quick test

celenechang: removed! thanks for catching

clamoriniere (approved):
👍  thank you for this fix

PR #80, opened 2021-01-22
Fix missing pause reason in status

clamoriniere:
### What does this PR do?

* Add more `ExtendedDaemonSetStatusReason`
* Check also Init and Ephemeral container state

### Motivation

### Additional Notes

### Describe your test plan

### DIFF
--- a/pkg/controller/utils/pod/pod.go
+++ b/pkg/controller/utils/pod/pod.go
@@ -183,7 +183,8 @@ func convertReasonToEDSStatusReason(reason string) datadoghqv1alpha1.ExtendedDae
 		datadoghqv1alpha1.ExtendedDaemonSetStatusReasonCreateContainerConfigError,
 		datadoghqv1alpha1.ExtendedDaemonSetStatusReasonCreateContainerError,
 		datadoghqv1alpha1.ExtendedDaemonSetStatusReasonPreStartHookError,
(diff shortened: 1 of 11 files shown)


pkg/controller/utils/pod/pod.go
```diff
@@ -158,17 +163,44 @@ func IsCannotStartReason(reason string) bool {
...
+	for _, s := range containerStatusList(pod) {
 		if s.State.Waiting != nil && IsCannotStartReason(s.State.Waiting.Reason) {
-			return true, datadoghqv1alpha1.ExtendedDaemonSetStatusReason(s.State.Waiting.Reason)
+			return true, convertReasonToEDSStatusReason(s.State.Waiting.Reason)
 		}
 	}
 	return false, datadoghqv1alpha1.ExtendedDaemonSetStatusReasonUnknown
 }
 
+func convertReasonToEDSStatusReason(reason string) datadoghqv1alpha1.ExtendedDaemonSetStatusReason {
```
xornivore, line 174: This is 1:1 by most part, replace with a string to bool map lookup maybe?

xornivore (approved):
Small nit.

pkg/controller/utils/pod/pod.go
clamoriniere: @xornivore what do you think about my change?

xornivore (approved):

PR #84, opened 2021-03-29
Support pausing and freezing non-canary replicasets

ahmed-mez:
### What does this PR do?

- Support pausing rolling updates.
  - kubectl plugin commands:
    - `k eds pause-rolling-update <eds name>` 
    - `k eds unpause-rolling-update <eds name>` 
  - When paused:
    - The replicaset controller won't delete pods anymore which prevents the rolling update from proceeding
    - The controller won't block the creation of new pods, so if new nodes appear in the meantime the controller will create new pods on the new nodes even when paused
    - Pausing a rolling update when the ERS is deployed for the first time in the cluster will not have any effect (just like canary)
    - The metric `eds_status_rolling_update_paused` is set to `1`

- Support freezing rollouts.
  - kubectl plugin commands:
    - `k eds freeze-rollout <eds name>` 
    - `k eds unfreeze-rollout <eds name>` 
  - When frozen:
    - The replicaset controller won't delete nor create new pods
    - The controller will block the creation of new pods, so if new nodes appear in the meantime the controller will not create new pods
    - Freezing a rollout will take effect even if the ERS is deployed for the first time in the cluster
    - The metric `eds_status_rollout_frozen` is set to `1`
 
### Motivation

More flexibility and better control of the EDS when deploying new changes

### Additional Notes

Make sure you build the kubectl plugin locally during QA `make kubectl-eds` -> `./bin/kubectl-eds`

### Describe your test plan

- Make sure the described behaviour above is respected
- Make sure the EDS state is updated accordingly in each case
- Make sure the ERS conditions are updated accordingly in each case
- Make sure the metrics are updated accordingly
- Make sure the kubectl plugin commands don't succeed when in canary phase
- Try edge cases like freezing and pausing at the same time, make sure the freeze wins
- Be imaginative and break the feature! 🔨

### DIFF
--- a/controllers/extendeddaemonset_test.go
+++ b/controllers/extendeddaemonset_test.go
@@ -14,11 +14,15 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	datadoghqv1alpha1 "github.com/DataDog/extendeddaemonset/api/v1alpha1"
+	"github.com/DataDog/extendeddaemonset/controllers/extendeddaemonsetreplicaset/conditions"
+	"github.com/DataDog/extendeddaemonset/controllers/extendeddaemonsetreplicaset/strategy"
 	"github.com/DataDog/extendeddaemonset/controllers/testutils"
 	// +kubebuilder:scaffold:imports
 )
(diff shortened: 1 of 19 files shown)


controllers/extendeddaemonset_test.go
```diff
@@ -85,3 +89,375 @@ var _ = Describe("ExtendedDaemonSet Controller", func() {
...
+					"ERS status should be updated: %#v",
+					ers.Status,
+				)
+			},
+			)
+		})
+	})
+})
+
+var _ = Describe("ExtendedDaemonSet Rollout Freeze", func() {
```
clamoriniere, line 292: 👏  for the tests

controllers/extendeddaemonsetreplicaset/strategy/rollingupdate.go
```diff
@@ -117,8 +125,17 @@ func ManageDeployment(client runtimeclient.Client, params *Parameters) (*Result,
 	nbPodToCreateWithConstraint := utils.MinInt(nbPodToCreate, len(allPodToCreate))
 	params.Logger.V(1).Info("Pods actions with limits", "nbPodToDelete", nbPodToDelete, "nbPodToCreate", nbPodToCreate, "nbPodToDeleteWithConstraint", nbPodToDeleteWithConstraint, "nbPodToCreateWithConstraint", nbPodToCreateWithConstraint)
```
clamoriniere, line 126: could you update the log line with the `frozen` and `paused` info?

clamoriniere (commented):
👏 for the PR.
Just a small comment about a log line.

controllers/extendeddaemonsetreplicaset/strategy/rollingupdate.go
ahmed-mez: good point! done

clamoriniere (approved):

PR #86, opened 2021-04-08
Fix how EDS is using the kube-scheduler when possible

clamoriniere:
### What does this PR do?

* Fix how the ExtendedDaemonset controller is setting the `pod.Spec.NodeName` even if the `Node` affinity is used.
* Unscheduled Pod are now also assign to Node.

### Motivation

Depending of the kubernetes version (<= 1.12), the Daemonset controller is not using the kube-scheduler to schedule a Pod on a specific Node. Instead it set directly the `nodeName` in the pod spec.
Extendeddaemonset is doing the same thing. But a bug was still setting the `nodeName` even if "node affinity" was used.

### Additional Notes

### DIFF
--- a/controllers/extendeddaemonset_test.go
+++ b/controllers/extendeddaemonset_test.go
@@ -44,7 +44,7 @@ var _ = Describe("ExtendedDaemonSet Controller", func() {
 			Name:      name,
 		}
 
-		It("Should create one pod by nodes", func() {
+		It("Should create one pod by node", func() {
 			edsOptions := &testutils.NewExtendedDaemonsetOptions{
 				CanaryStrategy: nil,
 				RollingUpdate: &datadoghqv1alpha1.ExtendedDaemonSetSpecStrategyRollingUpdate{
--- a/controllers/extendeddaemonset_test.go
+++ b/controllers/extendeddaemonset_test.go
@@ -44,7 +44,7 @@ var _ = Describe("ExtendedDaemonSet Controller", func() {
 			Name:      name,
 		}
 
-		It("Should create one pod by nods", func() {
+		It("Should create one pod by nodes", func() {
 			edsOptions := &testutils.NewExtendedDaemonsetOptions{
 				CanaryStrategy: nil,
 				RollingUpdate: &datadoghqv1alpha1.ExtendedDaemonSetSpecStrategyRollingUpdate{
--- a/controllers/extendeddaemonsetreplicaset/filters.go
+++ b/controllers/extendeddaemonsetreplicaset/filters.go
@@ -56,12 +56,11 @@ func FilterAndMapPodsByNode(logger logr.Logger, replicaset *datadoghqv1alpha1.Ex
 	}
 
 	for id, pod := range podList.Items {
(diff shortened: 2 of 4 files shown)


controllers/extendeddaemonsetreplicaset/filters.go
```diff
@@ -71,9 +70,15 @@ func FilterAndMapPodsByNode(logger logr.Logger, replicaset *datadoghqv1alpha1.Ex
 				continue
 			}
 
-			podsByNodeName[pod.Spec.NodeName] = append(podsByNodeName[pod.Spec.NodeName], &podList.Items[id])
+			podsByNodeName[nodeName] = append(podsByNodeName[nodeName], &podList.Items[id])
+
+			if _, scheduled := podutils.IsPodScheduled(&pod); !scheduled {
+				unscheduledPods = append(unscheduledPods, &podList.Items[id])
+
+				continue
```
celenechang, line 78: is the `continue` needed?

controllers/extendeddaemonset_test.go
```diff
@@ -44,7 +44,7 @@ var _ = Describe("ExtendedDaemonSet Controller", func() {
 			Name:      name,
 		}
 
-		It("Should create one pod by nods", func() {
+		It("Should create one pod by nodes", func() {
```
celenechang, line 47: i think it would be better as `by node`

controllers/extendeddaemonsetreplicaset/filters.go
```diff
@@ -101,7 +106,11 @@ func FilterAndMapPodsByNode(logger logr.Logger, replicaset *datadoghqv1alpha1.Ex
 
 	// add duplicated pods to the pod deletion slice
 	for _, pod := range duplicatedPods {
-		logger.V(1).Info("PodToDelete", "reason", "duplicatedPod", "pod.Name", pod.Name, "node.Name", pod.Spec.NodeName)
+		nodeName, err := podutils.GetNodeNameFromPod(pod)
+		if err != nil {
+			continue
+		}
+		logger.V(1).Info("PodToDelete", "reason", "duplicatedPod", "pod.Name", pod.Name, "node.Name", nodeName)
```
celenechang, line 113: it looks like if we don't find the `nodeName`, we skip the log line but still delete the pod. is that intended?

controllers/extendeddaemonsetreplicaset/filters.go
clamoriniere: not really :)

controllers/extendeddaemonsetreplicaset/filters.go
clamoriniere: right, 
I think, this case should not happen, and it is fine if we delete it in that case.
I will update the code to remove the error check and print the log line in any case.

clamoriniere:
@celenechang thanks for the review. I have update the PR. Ready of a second review :)

celenechang (approved):
lgtm! can a unit test be added for `GetNodeNameFromPod` ?

PR #88, opened 2021-04-10
Delete failed pods with a backoff

celenechang:
### What does this PR do?

- Removes the effect of the `ignoreEvictedPods` flag
- Instead of handling/ignoring evicted pods specifically, delete all `failed` pods with an exponential backoff by `ExtendedDaemonsetReplicaset` and `Node`

### Motivation

Better handling of evicted (and all failed) pods

### Additional Notes

### Describe your test plan

- Start a `kind` cluster
- Build Docker image and deploy to cluster (with `make` commands)
- Apply an EDS spec with ephemeral storage requests and limits
- Create a file larger than the limit by running `bash -c 'dd if=/dev/urandom of=/noise bs=1M count=130'` on one of the pods
- The pod should be evicted and deleted right away. Run the `bash` command for the pod that replaces it (i.e. on the same node). Keep an eye on how long it takes for the `Evicted` pod to clear.
- After repeating a few times it should be clear that the `Evicted` pod takes longer to get deleted each time.
- Also try this on a Canary deployment

### DIFF
--- a/controllers/extendeddaemonsetreplicaset/controller.go
+++ b/controllers/extendeddaemonsetreplicaset/controller.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	utilserrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/flowcontrol"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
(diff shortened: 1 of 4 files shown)


controllers/extendeddaemonsetreplicaset/controller.go
```diff
@@ -163,6 +166,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	reqLogger.V(1).Info("Updating ExtendedDaemonSetReplicaSet status")
 	err = r.updateReplicaSet(replicaSetInstance, newStatus)
 
+	// Garbage collect the failedPodsBackOff map once per minute,
+	// i.e. whenever the seconds [0,60] is less than the reconcile frequency
+	if now.Time.Second() < int(daemonsetInstance.Spec.Strategy.ReconcileFrequency.Duration.Seconds()) {
+		r.failedPodsBackOff.GC()
```
clamoriniere, line 172: could you add a comment about the `GC()` behaviour? 
```suggestion
		// Garbage collect records that have aged past maxDuration (here: 15min).
		r.failedPodsBackOff.GC()
```

clamoriniere (approved):
💯 
Thanks for implementing this backoff policy.
just a small nit to give more details about how `GC()` works.

PR #98, opened 2021-06-10
Add completion pct target to eds-check

vboulineau:
### What does this PR do?

Allows to configure completion pct based on env vars, default to 95% or < 10 nodes

### Motivation

### Additional Notes

### Describe your test plan

`eds-check` should return true as soon as 95% of nodes are done (or less than 10 nodes are remaining in case of small clusters)

### DIFF
--- a/cmd/check-eds/upgrade/upgrade.go
+++ b/cmd/check-eds/upgrade/upgrade.go
@@ -9,6 +9,8 @@ package upgrade
 import (
 	"context"
 	"fmt"
+	"os"
+	"strconv"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


cmd/check-eds/upgrade/upgrade.go
```diff
@@ -137,14 +157,15 @@ func (o *Options) Run() error {
...
 		}
-		if eds.Status.UpToDate < eds.Status.Current {
-			o.printOutf("still upgrading nb pods: %d, nb updated pods: %d", eds.Status.Current, eds.Status.UpToDate)
 
-			return false, nil
+		if float64(eds.Status.UpToDate)*o.nodeCompletionPct > float64(eds.Status.Current) ||
+			eds.Status.Current-eds.Status.UpToDate < o.nodeCompletionMin {
+			o.printOutf("upgrade is now finished (reached threshold): %d, nb updated pods: %d, threshold pct: %f, min threshold: %d", eds.Status.Current, eds.Status.UpToDate, o.nodeCompletionPct, o.nodeCompletionMin)
+
+			return true, nil
```
clamoriniere, line 165: could you add a log if `false` to still see that the process is running?

clamoriniere (approved):

PR #107, opened 2021-07-26
Automate EDS Krew plugin releasing

clamoriniere:
### What does this PR do?

Use a new github action to automatically create a PR on `krew-index`

### Motivation

Less action done manually during a release

### Additional Notes

### Describe your test plan


.github/workflows/release.yaml
```diff
@@ -2,8 +2,11 @@ name: release
 
 on:
   push:
+    branches:
+      - main
+      - "v[0-9]+.[0-9]+"
     tags:
-      - 'v*'
+      - "v*"
```
davidor, line 9: This will also publish release candidates. Is that OK or should we change it?

clamoriniere: I wondering what is the purpose of using `"v*"` instead of `'v*'` ?

yes, we are publicly publishing RC. to ease QA on public environment and also to allow external contributors/users to test release candidate.

davidor: I don't know if simple vs double quotes make a difference in this particular case. I guess they're equivalent.

davidor (approved):

PR #108, opened 2021-08-03
Adapt metadata validation patch for CRD v1

davidor:
### What does this PR do?

2 of the 3 templates in `config/crd/bases/v1` fail when applying them with this error:
```
unknown field "validation" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec
```

That `validation` field is added by `hack/patch-crds.sh`. It works with `v1beta1`, but not with `v1`.

This PR fixes the issue by adapting `hack/patch-crds.sh` and the file it uses (`patch-crd-metadata.yaml`) to work with `v1`.

### Describe your test plan

Run `kubectl apply -f config/crd/bases/v1/`. It should not raise any errors.

### DIFF
--- a/config/crd/bases/v1/datadoghq.com_extendeddaemonsetreplicasets.yaml
+++ b/config/crd/bases/v1/datadoghq.com_extendeddaemonsetreplicasets.yaml
@@ -118,6 +118,131 @@ spec:
                   metadata:
                     description: 'Standard object''s metadata. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata'
                     type: object
+                    properties:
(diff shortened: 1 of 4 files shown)


davidor:
I've marked this as a draft PR because I'm not sure whether I should update the manifests in the `bundle` and the `chart` dirs manually now or if it's done only when there's a release.

davidor:
I've seen that the `bundle` dir only contains the v1beta1 CRDs, so this PR does not affect what's in there. Not sure if v1 CRDs should be part of the bundle, though.

clamoriniere (approved):

PR #110, opened 2021-08-06
Generate a PodTemplate for each EDS

ahmed-mez:
### What does this PR do?

Generate a PodTemplate for each EDS as preparation for https://github.com/kubernetes/autoscaler/pull/3964

### Motivation

Read https://github.com/kubernetes/autoscaler/issues/3873 for context

### Additional Notes

### Describe your test plan

Deploy a `EDS` and make sure a corresponding `PodTemplate` object with the same name and namespace has the same lifecycle of the `EDS` object (creation/update/deletion) and has the label `cluster-autoscaler.kubernetes.io/daemonset-pod=true`


clamoriniere (approved):
💯

PR #116, opened 2021-08-12
add tests in pkg/controller/utils

celenechang:
### What does this PR do?

Increase test coverage

### Motivation

### Additional Notes

### Describe your test plan

### DIFF
--- a/pkg/controller/utils/comparison/comparison_test.go
+++ b/pkg/controller/utils/comparison/comparison_test.go
@@ -149,7 +149,7 @@ func Test_StringsContains(t *testing.T) {
 			want: false,
 		},
 		{
-			name: "a does not contain x v2",
+			name: "a does not contain x (but it is a substring of a string in a)",
 			a: []string{
 				"hello",
 				"goodbye",
--- a/pkg/controller/utils/comparison/comparison_test.go
+++ b/pkg/controller/utils/comparison/comparison_test.go
@@ -5,7 +5,15 @@
 
 package comparison
 
-import "testing"
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	datadoghqv1alpha1 "github.com/DataDog/extendeddaemonset/api/v1alpha1"
+)
 
 func TestGenerateHashFromEDSResourceNodeAnnotation(t *testing.T) {
 	type args struct {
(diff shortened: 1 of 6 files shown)


pkg/controller/utils/comparison/comparison_test.go
```diff
@@ -66,3 +74,94 @@ func TestGenerateHashFromEDSResourceNodeAnnotation(t *testing.T) {
...
+			name: "a does not contain x",
+			a: []string{
+				"hello",
+				"goodbye",
+			},
+			x:    "hi",
+			want: false,
+		},
+		{
+			name: "a does not contain x v2",
```
davidor, line 152: What's the difference with the previous test?

pkg/controller/utils/list_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+package utils
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestContainsString(t *testing.T) {
```
davidor, line 14: Not something introduced in this PR, but the function under test does the same as the one you already tested in `func Test_StringsContains`. Should we get rid of one of them?

pkg/controller/utils/pod/create_test.go
```diff
@@ -129,3 +133,34 @@ func Test_overwriteResourcesFromNode(t *testing.T) {
...
+		},
+	}
+
+	templateCopy := templateOriginal.DeepCopy()
+	// nil case, no change to template
+	edsNode := &datadoghqv1alpha1.ExtendedDaemonsetSetting{}
+	overwriteResourcesFromEdsNode(templateOriginal, edsNode)
+	assert.Equal(t, templateCopy, templateOriginal)
+
+	resouresRef := corev1.ResourceList{
```
davidor, line 152: typo

pkg/controller/utils/pod/create_test.go
```diff
@@ -129,3 +133,34 @@ func Test_overwriteResourcesFromNode(t *testing.T) {
...
+	edsNode = datadoghqv1alpha1test.NewExtendedDaemonsetSetting("foo", "bar", "reference", &datadoghqv1alpha1test.NewExtendedDaemonsetSettingOptions{
+		CreationTime: time.Now(),
+		Resources: map[string]corev1.ResourceRequirements{
+			"container1": {
+				Requests: resouresRef,
+			},
+		},
+	})
+	overwriteResourcesFromEdsNode(templateOriginal, edsNode)
+	assert.NotEqual(t, templateCopy, templateOriginal)
```
davidor, line 165: Apart from checking that they're not equal, is it worth testing that what changed is really the changes expected?

pkg/controller/utils/pod/pod_test.go
```diff
@@ -18,7 +18,354 @@ import (
...
+			Type:   v1.PodReady,
+			Status: v1.ConditionTrue,
+		},
+	}
+	isReady = IsPodReady(pod2)
+	assert.True(t, isReady)
+}
+
+func TestIsCannotStartReason(t *testing.T) {
+	reason := "ErrImagePull"
```
davidor, line 220: Is it worth to check all the reasons? Something like (not tested):
```
allReasons := _get_from_code_or_define_here_
for _, reason := range allReasons {
    cannotStart := IsCannotStartReason(reason)
    assert.True(t, cannotStart)
}
```

pkg/controller/utils/pod/pod_test.go
```diff
@@ -18,7 +18,354 @@ import (
...
+	pod3 := ctrltest.NewPod("bar", "pod3", "node1", &ctrltest.NewPodOptions{
+		CreationTimestamp: metav1.NewTime(time3),
+	})
+	pods := []*v1.Pod{
+		pod2,
+		pod3,
+		pod1,
+	}
+	podList := SortPodByCreationTime(pods)
+	assert.Equal(t, pod3, podList[0])
```
davidor, line 363: This could also be done in a single assert. I think it might make it easier understanding the issue if the test fails.
```go
assert.Equal(t, []*v1.Pod{pod3, pod2, pod1}, podList)
```

pkg/controller/utils/comparison/comparison_test.go
celenechang: not much, just that the string being searched is a substring of one of the words

pkg/controller/utils/list_test.go
celenechang: right, i noticed that. we could put them in a `common` file, perhaps in a separate PR

pkg/controller/utils/comparison/comparison_test.go
davidor: I see. What do you think about including that in the name of the test to make its objective more clear?

celenechang: sure i can do that!

pkg/controller/utils/pod/create_test.go
celenechang: thanks for the catch! typos bug me :)

pkg/controller/utils/pod/pod_test.go
celenechang: good point, thanks

pkg/controller/utils/pod/pod_test.go
celenechang: i like that, thanks

davidor (approved):

PR #118, opened 2021-08-30
fix condition for github action release

clamoriniere:
### What does this PR do?

the previous condition doesn't work.

### Motivation

The previous configuration prend the release job to run properly.

### Additional Notes

### Describe your test plan

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,9 +3,10 @@ name: release
 on:
   push:
     branches:
-      - "v*.*"
+      - "v[0-9]+.[0-9]+"
     tags:
-      - "v*.*.**"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -63,5 +66,6 @@ jobs:
       - 
         name: Update new plugin version in krew-index
         uses: rajatjindal/krew-release-bot@v0.0.40
+        continue-on-error: true
```
davidor, line 69: Why is this option needed here?

clamoriniere: I discover today that the action return an error when it detects that it is a release candidate
Link to the error: https://github.com/DataDog/datadog-operator/runs/3459789129?check_suite_focus=true

davidor (approved):

PR #127, opened 2021-09-06
[controllers/eds] Fix "available" and "up to date" in `get eds`

davidor:
### What does this PR do?

Fixes the "available" and "up to date" values in the `kubectl get eds` output.

"Available" should equal the sum of "available" of the active replica and the canary. There was a case where this was not working. When the canary duration expired and it became the active replica, "available" didn't take into account the replicas of the old set that were still there.

This PR also fixes a bug in the "Up to date" values. When there's a canary, "Up to date" should equal the number of replicas in the canary, but the code was adding the current ones plus the canary.

Notice that the base branch of this PR is `v0.7`

### Describe your test plan

Set up a cluster with more than 1 worker node. You can use kind for that.

Deploy the example EDS: `kubectl apply -f examples/foo-eds_v1.yaml`.

Run `kubectl get eds --watch` and `kubectl get ers --watch`. And wait until all the desired replicas are available.

Deploy the v2 of the example `kubectl apply -f examples/foo-eds_v2.yaml` (Adjust the timeout so you don't have to wait for 5 minutes). Check these 2 things:
- While the canary has not become the active replica, check that the "up to date" value of `get eds` corresponds to the number of "current" replicas of the canary. If you have not modified the number of replicas in `examples/foo-eds_v2.yaml`, this should be 0 (at the start) and 1 later. It should go over 1 only after the canary becomes the active replica.
- When the time that you chose in the previous step (or the default of 5 min) passes, pay attention to the "available" column of `get eds`. The value should equal the number of available replicas of the canary (that now is the active) plus the number of available replicas of the old set. Both of those values are shown in `get ers`.

### DIFF
--- a/controllers/extendeddaemonset/controller.go
+++ b/controllers/extendeddaemonset/controller.go
@@ -481,7 +481,7 @@ func manageStatus(status *datadoghqv1alpha1.ExtendedDaemonSetStatus, upToDate *d
 			status.Canary = &datadoghqv1alpha1.ExtendedDaemonSetStatusCanary{}
 		}
 		status.Desired += upToDate.Status.Desired
-		status.UpToDate += upToDate.Status.Available
+		status.UpToDate = upToDate.Status.Current
 		status.IgnoredUnresponsiveNodes += upToDate.Status.IgnoredUnresponsiveNodes
 
 		if !isCanaryPaused {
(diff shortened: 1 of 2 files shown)


clamoriniere (approved):
💯

PR #131, opened 2021-11-19
feat: add dd profiling

ahmed-mez:
### What does this PR do?

Add datadog profiling

### Motivation

Better observability

### Additional Notes

- Public docs https://docs.datadoghq.com/tracing/profiler/enabling/go/
- The exit code refactor is because of https://go-critic.com/overview.html#exitAfterDefer-ref

### Describe your test plan

Profiles appear on the profiler UI (should enable `ddProfilingEnabled`)


main.go
```diff
@@ -61,26 +62,46 @@ func main() {
...
 
 	// Parsing flags
 	flag.Parse()
 
+	exitCode := 0
+	defer func() { os.Exit(exitCode) }()
+
+	if ddProfilingEnabled {
+		setupLog.Info("Starting datadog profiler")
+		if err := profiler.Start(
```
davidor, line 81: Would it be useful to add `profiler.WithService("eds")` or similar to be able to filter later?

main.go
```diff
@@ -61,26 +62,46 @@ func main() {
...
+		exitCode = 1
+
+		return
 	}
 
 	// Print version information
 	if printVersion {
 		version.PrintVersionWriter(os.Stdout)
-		os.Exit(0)
+		exitCode = 1
```
davidor, line 102: This shouldn't be an error.

main.go
ahmed-mez: This needs to stay configurable depending on how users want to name the service, for internal usage I'm planning to define `DD_SERVICE` in the eds pod, it will be read and applied by the profiler library.

main.go
ahmed-mez: yes nice catch!

davidor (approved):

PR #132, opened 2021-11-22
fix: simplify cleanupReplicaSet for less memory allocations

ahmed-mez:
### What does this PR do?

Simplify `cleanupReplicaSet`: instead of getting full pod list to decide whether a ERS must be deleted, rely on the info stored in its status.

Querying the a pod list is a costly operation especially in larger clusters during a rolling update as the old ERS still have a large non-nil list of pods attached to it.

### Motivation

image

A is before the rolling update, B is during the rolling update (cluster size: 1.6k nodes)

### Additional Notes

`cleanupReplicaSet` is covered by unit tests/

### Describe your test plan

Rollout new workload versions multiple times, make sure old ERSs get deleted.

### DIFF
--- a/controllers/extendeddaemonset/controller.go
+++ b/controllers/extendeddaemonset/controller.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"math/rand"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


clamoriniere (commented):
this is a cleaver solution to know when a ERS needs to be delete.
Since we do already the operation in the ERS controller it definitely makes sens to use the ers.status to get the info 💯

clamoriniere (approved):

PR #138, opened 2021-12-13
move failed metric from EDS to ERS

celenechang:
### What does this PR do?

The `canary_failed` metric was either not reporting, or reporting very sparsely (i.e. a single point).

This is because the metric depends on the EDS status to report, but upon canary failure the canary is cleared right away and the EDS status is updated accordingly.

Since the canary failure is inherently tied to the ERS, and the controller logic makes it tricky to keep the metric based on the EDS status, this PR changes this metric to report from the ERS instead. If the ERS needs to be removed because of a canary failure, there is a delay of 5m to give the metric that amount of time to report.

### Motivation

Improve telemetry so that canary failures can easily be monitored 

### Additional Notes

### Describe your test plan

Force a canary failure (made easier with subsequent PR #139)
After failure, but within 5m, check the ERS is still present: `k get ers`
Exec onto the EDS controller pod, and check that the failed metric is reporting

```
$ k exec -it eds-controller-manager-xxx -- bash
$ curl http://localhost:8080/ksmetrics | grep fail
# HELP ers_status_canary_failed The failed state of the canary deployment, set to 1 if failed, else 0
# TYPE ers_status_canary_failed gauge
ers_status_canary_failed{namespace="default",name="foo-4prqt"} 0
ers_status_canary_failed{namespace="default",name="foo-4nqsj"} 1
```

### DIFF
--- a/controllers/extendeddaemonset/controller.go
+++ b/controllers/extendeddaemonset/controller.go
@@ -28,6 +28,7 @@ import (
 
 	datadoghqv1alpha1 "github.com/DataDog/extendeddaemonset/api/v1alpha1"
 	"github.com/DataDog/extendeddaemonset/controllers/extendeddaemonset/conditions"
+	ersconditions "github.com/DataDog/extendeddaemonset/controllers/extendeddaemonsetreplicaset/conditions"
 	"github.com/DataDog/extendeddaemonset/controllers/extendeddaemonsetreplicaset/scheduler"
 	"github.com/DataDog/extendeddaemonset/pkg/controller/metrics"
 	"github.com/DataDog/extendeddaemonset/pkg/controller/utils"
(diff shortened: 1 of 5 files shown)


controllers/extendeddaemonset/controller.go
```diff
@@ -568,11 +569,20 @@ func (r *Reconciler) cleanupReplicaSet(logger logr.Logger, rsList *datadoghqv1al
...
 	}
 
+	// If canary deploy has failed, delay ERS deletion for 5m so that failed metric reports
+	if ersconditions.IsConditionTrue(&ers.Status, datadoghqv1alpha1.ConditionTypeCanaryFailed) {
+		failedIdx := ersconditions.GetIndexForConditionType(&ers.Status, datadoghqv1alpha1.ConditionTypeCanaryFailed)
+		t := ers.Status.Conditions[failedIdx].LastTransitionTime.Add(time.Minute * 5)
+		if now.Before(t) {
+			return false
+		}
+	}
```
ahmed-mez, line 584: I guess there are no side effects to this delay? 🤔  In any case, 5m seem a bit long if we consider the agent check interval which is 15s. 1 or 2 minutes should be enough to collect multiple data points.

ahmed-mez (approved):
LGTM, just a small question about the delay duration.

controllers/extendeddaemonset/controller.go
celenechang: no side effects! the active ERS reconciles as expected. i guess i went with 5m to compensate for the fact it is hard/impossible to see the points on the dashboard currently. but we can do 2m and see how that feels. thanks for the review!

PR #139, opened 2021-12-13
add generic maxDuration for autofailure

celenechang:
### What does this PR do?

Adds `canary.autoFail.maxDuration` so that a canary that is paused due to something other than restarts can eventually be automatically failed and the active ERS will be reset to the current one.

This value must be greater than the canary duration to work.

### Motivation

Better experience with automated canary features. In addition, ease of testing the autoFail path.

### Additional Notes

### Describe your test plan

Update the example EDS to include:
```
      autoFail:
        enabled: true
        maxDuration: 8m
```
and an error condition, such as a bad image name.

After the `maxDuration` fo the canary has passed, check that the canary is over and the active ERS has stayed the same.

Also, make sure that if `canary.autoFail.maxDuration < canary.duration` then the parameter is not respected. (Can look for the log line `Canary autoFail.maxDuration must be greater than the canary duration.` as well.)


controllers/extendeddaemonsetreplicaset/strategy/canary.go
```diff
@@ -154,6 +157,26 @@ func manageCanaryPodFailures(pods []*v1.Pod, params *Parameters, result *Result,
...
+		autoFailMaxDuration = canary.AutoFail.MaxDuration
+
+		// autoFailMaxDuration must be greater than canaryDuration
+		if autoFailMaxDuration.Duration <= canaryDuration.Duration {
+			params.Logger.Info("Canary autoFail.maxDuration must be greater than the canary duration.")
+			if autoPauseEnabled {
+				params.Logger.Info("With AutoPause enabled, canary.autoFail.maxDuration >= 2*canary.duration is recommended.")
+			}
+			autoFailMaxDuration = nil
+		}
```
ahmed-mez, line 174: I'm wondering if we should handle this in the CR validation phase, the info log here is useful but if the user doesn't see it will be a silent failure. Or at least we should log it as error and say explicitly that this config will be ignored.

api/v1alpha1/extendeddaemonset_types.go
```diff
@@ -99,22 +99,32 @@ type ExtendedDaemonSetSpecStrategyCanary struct {
...
 	Enabled *bool `json:"enabled,omitempty"`
 	// MaxRestarts defines the number of tolerable (per pod) Canary pod restarts after which the Canary deployment is autofailed.
+	// Default value is 5.
 	MaxRestarts *int32 `json:"maxRestarts,omitempty"`
 	// MaxRestartsDuration defines the maximum duration of tolerable Canary pod restarts after which the Canary deployment is autofailed.
+	// There is no default value.
 	MaxRestartsDuration *metav1.Duration `json:"maxRestartsDuration,omitempty"`
+	// MaxDuration defines the maximum duration of a Canary after which the Canary deployment is autofailed. This is a safeguard against lengthy Canary pauses.
+	// There is no default value.
+	MaxDuration *metav1.Duration `json:"maxDuration,omitempty"`
```
ahmed-mez, line 127:
```suggestion
	MaxCanaryDuration *metav1.Duration `json:"maxDuration,omitempty"`
```

No strong opinion, it's just that the other option `MaxRestartsDuration` makes `MaxDuration` a bit confusing. `MaxCanaryDuration` would clearer imho.

controllers/extendeddaemonsetreplicaset/strategy/canary_test.go
```diff
@@ -610,6 +621,111 @@ func TestManageCanaryStatus_LongRestartsDurationLeadingToFail(t *testing.T) {
 	test.Run(t)
 }
 
+func TestManageCanaryStatus_LongDurationLeadingToFail(t *testing.T) {
+	now := time.Now()
+	imageErrorStartedAt := now.Add(-time.Hour)
+	// restartsUpdatedAt := now.Add(-10 * time.Minute)
+	// restartedAt := now.Add(-time.Minute)
```
ahmed-mez, line 628: ✂️

celenechang: 🙈

controllers/extendeddaemonsetreplicaset/strategy/canary.go
celenechang: yes, i like this! just noticed extendeddaemonset_validate.go , i will move it to there

ahmed-mez (approved):

PR #141, opened 2022-01-04
change fail cmd to update ers instead of eds

celenechang:
### What does this PR do?

- Update plugin `canary fail` cmd to update the status of the ERS instead of using EDS annotations.
- Remove references to canary fail annotations that are no longer needed
- Remove other methods that are not used
- Add TODOs for future PRs
 
### Motivation

It was noticed that the behavior of the `canary fail` command was peculiar; manual failures were not leading to canary pod deletions as expected. This seemed to be caused by the canary status being failed too quickly before the failed annotation could trigger the appropriate action.

Since failures are inherently tied to an ERS, it makes more sense to have the plugin modify the ERS instead. This more closely mimics the behavior for auto-failures and removes an extra step in managing the canary.

### Additional Notes

### Describe your test plan

- Build the plugin with `make kubectl-eds`
- Apply the example extended daemonset `foo` (`k apply -f examples/foo-eds_v1.yaml`)
- Apply another extended daemonset to trigger the canary (`k apply -f examples/foo-eds_v2.yaml`)
- Fail the canary with `bin/kubectl-eds canary fail foo`
- Make sure that the failed ERS status conditions are updated to show
```
    Reason:                    Manually failed
    Status:                    True
    Type:                      Canary-Failed
```
- Make sure that the EDS status conditions are updated to show
```
    Message:                   canary failed with ers: foo-5wgn2
    Reason:                    CanaryFailed
    Status:                    False
    Type:                      Canary-Failed
```
- Make sure that the failed canary pods are cleared

### DIFF
--- a/controllers/extendeddaemonsetreplicaset/controller.go
+++ b/controllers/extendeddaemonsetreplicaset/controller.go
@@ -223,11 +223,6 @@ func (r *Reconciler) applyStrategy(logger logr.Logger, daemonset *datadoghqv1alp
 		conditions.UpdateExtendedDaemonSetReplicaSetStatusCondition(strategyParams.NewStatus, now, datadoghqv1alpha1.ConditionTypeActive, corev1.ConditionFalse, "", "", false, false)
 		logger.Info("manage canary deployment")
 		strategyResult, err = strategy.ManageCanaryDeployment(r.client, daemonset, strategyParams)
-	case strategy.ReplicaSetStatusCanaryFailed:
-		logger.Info("manage canary failed deployment")
-		strategyResult = &strategy.Result{
-			NewStatus: strategyParams.NewStatus.DeepCopy(),
-		}
 	case strategy.ReplicaSetStatusUnknown:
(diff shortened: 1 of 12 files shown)


controllers/extendeddaemonsetreplicaset/controller.go
```diff
@@ -223,11 +223,6 @@ func (r *Reconciler) applyStrategy(logger logr.Logger, daemonset *datadoghqv1alp
 		conditions.UpdateExtendedDaemonSetReplicaSetStatusCondition(strategyParams.NewStatus, now, datadoghqv1alpha1.ConditionTypeActive, corev1.ConditionFalse, "", "", false, false)
 		logger.Info("manage canary deployment")
 		strategyResult, err = strategy.ManageCanaryDeployment(r.client, daemonset, strategyParams)
-	case strategy.ReplicaSetStatusCanaryFailed:
-		logger.Info("manage canary failed deployment")
-		strategyResult = &strategy.Result{
-			NewStatus: strategyParams.NewStatus.DeepCopy(),
-		}
```
celenechang, line 230: this was added in https://github.com/DataDog/extendeddaemonset/pull/66 but i don't think this case is ever visited

clamoriniere: right

clamoriniere (approved):
looks good 👍  and it will simplify the logic 👏

PR #142, opened 2022-01-06
add node taint check for canary nodes, add missing test case for validation

celenechang:
### What does this PR do?

During QA it was noticed that node taints are not handled properly when scheduling canary pods. This fixes the issue.

In addition, a test case is added that was missing for a recently added canary parameter (autoFail.canaryTimeout).

### Motivation

Fix

### Additional Notes

### Describe your test plan

Start a `kind` cluster with a tainted node (master nodes are tainted by default)
Deploy the EDS controller with `make deploy`
Deploy an example EDS that doesn't have any tolerations (edit examples/foo-eds_v1.yaml and then run `k apply -f examples/foo-eds_v1.yaml`)
Deploy another EDS to kick off the canary, also without any tolerations `examples/foo-eds_v2.yaml`
Make sure that the canary pod(s) gets scheduled

### DIFF
--- a/api/v1alpha1/extendeddaemonset_validate_test.go
+++ b/api/v1alpha1/extendeddaemonset_validate_test.go
@@ -7,6 +7,7 @@ package v1alpha1
 
 import (
 	"testing"
+	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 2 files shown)


clamoriniere (approved):
nice 💯

L3n41c (approved):

PR #150, opened 2022-07-01
Add builds for arm64 platforms

paulcacheux:
### What does this PR do?

This PR:
- adds support for arm64 platforms when releasing binaries
- adds a build workflow, running goreleaser build, which is useful to test PR of this kind

### Motivation

### Additional Notes

### Describe your test plan

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,22 @@
+name: build
+
+on:
+  push:
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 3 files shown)


paulcacheux:
Without the `--skip-validation` I'm getting
```
build failed after 0.04serror=git tag v0.7.0 was not made against commit 3b6505ef03bb01235daf0ca1544cad9e92abc12b
```
so `goreleaser` seems as confused with tags here as in the fork.

Maybe we can keep it, this way we split the concerns of releasing to the release workflow, and the concerns of building to this build workflow.

Otherwise I can simply remove this workflow and keep only the `arm64` change since we validated that the build would work.

CharlyF (approved):

PR #165, opened 2023-08-18
Fixed makefile, added support for ARM, and updated kustomization.yaml to use v1 apiextension version

adel121:
### What does this PR do?
- Fixes some issues in the Makefile (eg. misconfigured targets)
- Fixes the kubebuilder download link
- Fixes script file name
- Updates README.md

### Motivation
In the current state of the repository, several issues were noticed. Makefile fails to install CRDs into cluster. Reasons for this include:
- misconfigured targets
- typo in script name (underscore instead of hyphen)
- using version v1beta1 of apiextensions.k8s.io which is not supported in newer versions of kube clusters.
- uses an incorrect link to download kubebuilder
- uses a version of kubebuilder that doesn't support darwin_arm64 OS and platform

Additionally, the README.md was not fully updated after the release of version 0.8, resulting in the following discrepency:
- The README.md indicates that the only difference between `examples/foo-eds_v1.yaml` and `examples/foo-eds_v2.yaml` is the image tag used in the pod template. However, in the current state, in addition to the mentioned difference, one of the files configures `autoFail` in the canary strategy while the other one configures `autoPause`.

### Additional Notes
- The helm chart of the extended-daemon-set still uses v0.7 of EDS. The following PR upgrades the chart to use v0.8 of EDS.

### Describe your test plan
To test the changes, follow the steps in the README.md file.
- After creating a cluster of 3 nodes using kind and the configuration file provided in the repo, run `make install` and ensure the CRDs are created in the cluster without any error.
- Deploy and run the controller in the cluster using `make deploy`
- Try creating the resources defined in `examples/`
- Run `kubectl get pods`, the output should be similar to the following
image
- Try uninstalling the CRDs using `make uninstall`


fanny-jiang (approved):
💯

PR #166, opened 2023-08-29
Keep track of pods already unhealthy before deployment

adel121:
### What does this PR do?

This PR enhances the extended daemon set to keep track of pods that were already unhealthy before deployment.

[Jira Card](https://datadoghq.atlassian.net/jira/software/projects/CONT/boards/167?selectedIssue=CONT-4075)

### Motivation

The EDS controller like the Daemonset controller is configured to limit the number of daemonset pod un-ready during a rollout. For example the rollout strategy can be not more than 5% of pod unready. 
But if before a rollout start 4% of the existing pods are not ready, it means that the controller can also update 1% at the same time (5%-4%= 1%) which slow down the deployment. and even in some case block the deployment (if more than 5% of the pods are unready before the rollout).

If the EDS could keep track of the number of pods that were initially unhealthy before deployment, it would be able to keep the rollout strategy unaffected (and thus not blocked or slowed-down) by the pods that were previously unhealthy.

### Changes Made
- The number of unhealthy pods is added to the extendeddaemonset replicaset upon creation as an annotation `unready-pods`.
-  When calculating the number of pods that can be deleted in `limits.go`, the number of initially unhealthy pods (which is stored in the annotations of the ERS) is used to compensate for this difference in the calculation.

### Additional Notes
- The image `k8s.gcr.io/pause:latest` was used in the integration tests. However this image doesn't work on arm architecture, this is why I changed it to `k8s.gcr.io/pause:3.0`.
- In case the unready pods annotation doesn't exist for some reason, the reconciler will fallback to the default behaviour (i.e the rollout will be slowed down / blocked in case there were many pods unready before rollout)

### Describe your test plan
- Unit Testing:
    - The function `CalculatePodToCreateAndDelete` in `limits.go` was unit-tested in `limits_test.go` to ensure that the pods that were already unhealthy are taken into consideration.
    - We can run the unit tests by executing `make test`
- E2E Testing:
    - Integration testing was implemented
    - Deploy a kind cluster with a few nodes
    - Run the E2E test using `make e2e`
    - The E2E test of this change includes the following:
       - Creating an EDS with a corrupted image
       - This results in all pods of the EDS being unhealthy
       - Edit the EDS by setting a valid container image in the pod template (e.g. `k8s.gcr.io/pause:3.0`)
       - Launch the deployment
       - Ensure that the rollout is not blocked and that the EDS becomes finally in the Running state, with 1 active ERS and all pods are healthy and running
- QA Testing:
    - Create a kind cluster with few nodes (at least 3)
    - First create a configmap by running `kubectl create configmap example-configmap`
    - Run `make install` to install the CRDs
    - Run the controller (e.g by running `make run`, attention you should add `LeaderElectionNamespace: "kube-system"` to the created manager in `main.go` for this to work)
    - Create an EDS with the following description:
        ```apiVersion: datadoghq.com/v1alpha1
              kind: ExtendedDaemonSet
              metadata:
                   name: foo
              spec:
                strategy:
                  canary:
                    replicas: 1
                    duration: 5m
                    autoFail:
                      enabled: true
                      canaryTimeout: 10m
                    autoPause:
                      enabled: true
                      maxRestarts: 5
                  rollingUpdate:
                    maxParallelPodCreation: 1
                    maxUnavailable: 2
                    slowStartIntervalDuration: 1m
                template:
                  spec:
                    containers:
                      - name: daemon
                        image: k8s.gcr.io/pause:3.0
                        envFrom:
                          - configMapRef:
                              name: example-configmap
                    tolerations:
                      - operator: Exists
    - When the pods are up and running, delete the configmap by running `kubectl delete configmap example-configmap`
    - Delete some pods (at least 3)
    - The ERS will try to recreate the pods, but the created pods will become unhealthy because the configmap is deleted
    - Launch a deployment using the same definition of the EDS, but remove the `envFrom` property from the pod template
   - Check the status of the deployment process by running `kubectl get ers --watch`. You should see 1 canary and the active replicaset
   - After 5 minutes, the rollout should start and the deployment should not be blocked by the pods that are already unhealthy.
   - You should end up with 1 active ERS having all pods healthy and running

### DIFF
--- a/controllers/extendeddaemonset_e2e_test.go
+++ b/controllers/extendeddaemonset_e2e_test.go
@@ -95,7 +95,7 @@ var _ = Describe("ExtendedDaemonSet e2e updates and recovery", func() {
 				},
 			}
 
-			eds := testutils.NewExtendedDaemonset(namespace, name, "k8s.gcr.io/pause:latest", edsOptions)
+			eds := testutils.NewExtendedDaemonset(namespace, name, "registry.k8s.io/pause:3.0", edsOptions)
 			Expect(k8sClient.Create(ctx, eds)).Should(Succeed())
 
 			eds = &datadoghqv1alpha1.ExtendedDaemonSet{}
@@ -238,6 +238,7 @@ var _ = Describe("ExtendedDaemonSet e2e updates and recovery", func() {
 				return len(pods.Items) == 0
 			}), longTimeout, interval).Should(BeTrue(), "All EDS pods should be destroyed")
 		})
+
 	})
 })
 
@@ -271,7 +272,7 @@ var _ = Describe("ExtendedDaemonSet e2e PodCannotStart condition", func() {
 			},
 		}
 
-		eds := testutils.NewExtendedDaemonset(namespace, name, "k8s.gcr.io/pause:latest", edsOptions)
+		eds := testutils.NewExtendedDaemonset(namespace, name, "registry.k8s.io/pause:3.0", edsOptions)
 		Expect(k8sClient.Create(ctx, eds)).Should(Succeed())
 
 		eds = &datadoghqv1alpha1.ExtendedDaemonSet{}
@@ -465,7 +466,7 @@ var _ = Describe("ExtendedDaemonSet e2e successful canary deployment update", fu
 				},
 			}
 
-			eds := testutils.NewExtendedDaemonset(namespace, name, "k8s.gcr.io/pause:latest", edsOptions)
+			eds := testutils.NewExtendedDaemonset(namespace, name, "registry.k8s.io/pause:3.0", edsOptions)
 			Expect(k8sClient.Create(ctx, eds)).Should(Succeed())
 
 			eds = &datadoghqv1alpha1.ExtendedDaemonSet{}
@@ -486,7 +487,7 @@ var _ = Describe("ExtendedDaemonSet e2e successful canary deployment update", fu
 
 		It("Should do canary deployment", func() {
 			updateFunc := func(eds *datadoghqv1alpha1.ExtendedDaemonSet) {
-				eds.Spec.Template.Spec.Containers[0].Image = fmt.Sprintf("k8s.gcr.io/pause:3.1")
+				eds.Spec.Template.Spec.Containers[0].Image = fmt.Sprintf("registry.k8s.io/pause:3.1")
 			}
 
 			Eventually(updateEDS(k8sClient, key, updateFunc), timeout, interval).Should(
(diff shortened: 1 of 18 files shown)


api/v1alpha1/test/new.go
```diff
@@ -94,10 +94,12 @@ func NewExtendedDaemonSetReplicaSet(ns, name string, options *NewExtendedDaemonS
...
 		ObjectMeta: metav1.ObjectMeta{
-			Namespace:   ns,
-			Name:        name,
-			Labels:      map[string]string{},
-			Annotations: map[string]string{},
+			Namespace: ns,
+			Name:      name,
+			Labels:    map[string]string{},
+			Annotations: map[string]string{
+				"unready-pods": "0",
```
clamoriniere, line 101: we could define a more descriptive annotation key like `eds.datadoghq.com/unready-pods`

clamoriniere: Also this annotation key can be defined as `const string` to be re-use in other part of the code

controllers/extendeddaemonset_e2e_test.go
```diff
@@ -465,7 +466,7 @@ var _ = Describe("ExtendedDaemonSet e2e successful canary deployment update", fu
 				},
 			}
 
-			eds := testutils.NewExtendedDaemonset(namespace, name, "k8s.gcr.io/pause:latest", edsOptions)
+			eds := testutils.NewExtendedDaemonset(namespace, name, "k8s.gcr.io/pause:3.0", edsOptions)
```
clamoriniere, line 469: I think you can use the new kubernetes registry image

```
➜ crane ls  registry.k8s.io/pause
0.8.0
1.0
2.0
3.0
3.1
3.2
3.3
3.4.1
3.5
3.6
3.7
3.8
3.9
```

clamoriniere: but not `latest` because the tag is not updated

clamoriniere (commented):
Added some small comment, otherwise the logic implemented looks good 👍

api/v1alpha1/test/new.go
adel121: I agree. Thanks for suggesting this.

I will update the code.

controllers/extendeddaemonset_e2e_test.go
adel121: Alright. Will update the used image. Thank you

api/v1alpha1/test/new.go
```diff
@@ -94,10 +94,12 @@ func NewExtendedDaemonSetReplicaSet(ns, name string, options *NewExtendedDaemonS
...
 		ObjectMeta: metav1.ObjectMeta{
-			Namespace:   ns,
-			Name:        name,
-			Labels:      map[string]string{},
-			Annotations: map[string]string{},
+			Namespace: ns,
+			Name:      name,
+			Labels:    map[string]string{},
+			Annotations: map[string]string{
+				datadoghqv1alpha1.ExtendedDaemonSetReplicaSetUnreadyPodsAnnotationKey: "0",
```
clamoriniere, line 101: I think it is better to keep the `Annotations` by default empty, but us the `NewExtendedDaemonSetReplicaSetOptions` to add the annotations in some test cases.

I will be better to validate that the logic still work even if the annotation is not present.

adel121: Makes sense.

I updated the code. I had to default the number of unready pods to 0 in case the annotation didn't exist.

clamoriniere (approved):

PR #169, opened 2023-10-30
Update container start error so it respects `MaxSlowStartDuration`

jennchenn:
### What does this PR do?

During canary, if an error occurs during pod creation within the `MaxSlowStartDuration`, the pod is no longer auto-paused. If `MaxSlowStartDuration` is not defined, the behavior is unchanged.

### Motivation

Many recent deployments have been paused due to a `CreateContainerConfigError` - `Error: failed to sync secret cache: timed out waiting for the condition`. Though this error is sometimes rectified a few seconds/minutes after it is first raised, the deployment remains paused, thus slowing down releases, etc. This change is being introduced in hopes of avoiding cases where canary is paused although the issue no longer exists.

### Additional Notes

Should a default value for `maxSlowStartDuration` be set for our clusters? 

### Describe your test plan

Unit tests and E2E test were included with the change. I also deployed these changes to a few staging clusters and did not see the issue of a paused canary arising in those clusters.

### DIFF
--- a/controllers/extendeddaemonset_e2e_test.go
+++ b/controllers/extendeddaemonset_e2e_test.go
@@ -953,7 +953,7 @@ var _ = Describe("ExtendedDaemonSet e2e Pod within MaxSlowStartDuration", func()
 		}
 		Eventually(withList(listOptions, pods, "EDS pods", func() bool {
 			return len(pods.Items) == 0
-		}), timeout, interval).Should(BeTrue(), "All EDS pods should be destroyed")
+		}), longTimeout, interval).Should(BeTrue(), "All EDS pods should be destroyed")
 
 		erslist := &datadoghqv1alpha1.ExtendedDaemonSetReplicaSetList{}
 		Eventually(withList(listOptions, erslist, "ERS instances", func() bool {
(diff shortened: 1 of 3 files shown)


controllers/extendeddaemonset_e2e_test.go
```diff
@@ -893,6 +894,160 @@ var _ = Describe("ExtendedDaemonSet e2e rollout not blocked due to already faili
...
+
+		pods := &corev1.PodList{}
+		listOptions := []client.ListOption{
+			client.InNamespace(namespace),
+			client.MatchingLabels{
+				datadoghqv1alpha1.ExtendedDaemonSetNameLabelKey: name,
+			},
+		}
+		Eventually(withList(listOptions, pods, "EDS pods", func() bool {
+			if len(pods.Items) < 0 {
```
adel121, line 1001: ❓ 's
- Why would `len(pods.Items)` be negative in the first place? Doesn't the `len()` function always return non-negative value?
- What if `len(pods.Items) == 0` ? Should we return `false` in this case too?

adel121 (commented):
Thanks @jennchenn , seems good.

Left 2 small questions.

controllers/extendeddaemonset_e2e_test.go
jennchenn: meant to make this `== 0` thanks for catching that!

adel121 (approved):

celenechang (approved):

PR #196, opened 2024-12-11
Create go module for the CRD

davidor:
### What does this PR do?

Creates a new go module for the CRD.
This will allow other projects like the datadog-operator to rely on that go mod instead of the root go mod of the project.
This will reduce the dependencies required and hopefully create less conflicts when upgrading kubernetes libraries.

### Describe your test plan

I don't think there's anything specific apart from verifying that this simplifies the process when we need to upgrade the kubernetes libraries in the datadog-operator.


swang392 (approved):

davidor:
Rebased on main

PR #219, opened 2025-09-08
Use recommended GetScaledValue and fix canary computation when using percentage

tbavelier:
### What does this PR do?

* Uses `GetScaledValue` instead of `GetValue` as the latter was deprecated due to bug: https://pkg.go.dev/k8s.io/apimachinery@v0.28.9/pkg/util/intstr#GetValueFromIntOrPercent
* Uses EDS desired instead of ERS desired, the latter is computed after for canary, so when we use a %, we get % of 0 which is always 0, making the canary phase be skipped

### Motivation

* Fix internal deployment

### Additional Notes

### Describe your test plan

* Tested by using 3% in a 3 nodes cluster and ensuring I get a single canary replica

### DIFF
--- a/controllers/extendeddaemonset/controller.go
+++ b/controllers/extendeddaemonset/controller.go
@@ -256,7 +256,7 @@ func (r *Reconciler) updateInstanceWithCurrentRS(logger logr.Logger, now time.Ti
 
 		if isCanaryActive {
 			// manager CanaryNode selection.
-			nbCanaryPod, err := intstrutil.GetValueFromIntOrPercent(daemonset.Spec.Strategy.Canary.Replicas, int(daemonset.Status.Desired), true)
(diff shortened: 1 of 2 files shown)


controllers/extendeddaemonset/controller.go
```diff
@@ -361,7 +361,7 @@ func (r *Reconciler) selectNodes(logger logr.Logger, daemonset *datadoghqv1alpha
 		currentNodes = canaryStatus.Nodes
 	}
 
-	nbCanaryPod, err := intstrutil.GetValueFromIntOrPercent(daemonsetSpec.Strategy.Canary.Replicas, int(replicaset.Status.Desired), true)
+	nbCanaryPod, err := intstrutil.GetScaledValueFromIntOrPercent(daemonsetSpec.Strategy.Canary.Replicas, int(daemonset.Status.Desired), true)
```
tbavelier, line 364: Using EDS desired replicas is necessary here, as the canary ERS desired field is 0 at this point (so if using a percentage, we'd get `nbCanaryPod = 0` always)

